PR #115, opened 2019-03-27
switch to question mark

npfoss:
The star icon for things we aren't sure what they are is misleading I think. I replaced it with a question mark to hopefully indicate that we don't know, but this isn't a super satisfactory solution because these requirements will prevent the audit from saying you're good even if you fulfill them. A deep solution is probably not in the scope of MVP, but I want your input @wishdasher in particular (and anyone else) on what we should do about this. I think this PR is the minimum, but if it's urgent we can add more.


npfoss:
icons being handled better in #149

Issue #132, opened 2019-04-06
Update updateFulfillment function to use new progress routes on fireroad


georgiashay:
Now can log in and send roadID

PR #136, opened 2019-04-06
Progress overrides

### LINKED ISSUE #119: Add progress overrides
-Slider for plain-string requirements
-Updated at prefs/progress_overrides

georgiashay:
Able to change progress overrides through a dialog

Branches off from the progress-route branch

Closes #119 and #132


georgiashay:
* Fixed the error for General Mathematics, but there also seems to be an issue on the fireroad side with marking these types of requirements as completed, so they will not show as fulfilled right now
* It seems like the slider can't be made thicker using vuetify's exposed parameters, but I added a number input next to it which may be easier to use than the slider
* Progress overrides are not yet available for everything
* Icon problems -> #141 
* Added comments for list id.  Basically progress overrides is a dictionary where the keys are list-ids.  For sub requirements, the list ids are the parent's list id with dots and indices telling you where to find that requirement (ex: girs.0.2 = the CAL1 requirement).  So you need to know what list id you are at to update progress overrides.

georgiashay:
I’m working to change the FireRoad requirements (editor not working right now for some reason) but yeah it might be worth it to create an issue

The list-ids have to be dot delimited because that is what FireRoad expects.

PR #149, opened 2019-04-13
Icon things

### LINKED ISSUE #141: All the icon decisions
**Icons for plain string requirements:**
assignment_late for requirements with no manual progress set
assignment for requirements with a manual progress set (has a key in the progressOverrides dict)
assignment_turned_in for completed requirements (also make it green)

**Icons for regular requirements:**
Check (done) for completed
Dash (remove) for uncompleted

**Include icons in click event to open card**

**All requirements with progress bar**
Progress bar only green when full, otherwise yellow

**Info icons**
Only show on hover

**Close icons**
Courses on left
Everything else on top right

**Percentages show on hover**

georgiashay:
Closes #141 

Changes from description in #141:
-Added an orange color in addition to yellow on the progress bar (<15% complete)
-Can't access the dropdown icons, so instead of the tooltips, the percentage now displays next to the info icon (open to suggestions on better ways of displaying this, but I think it looks nice)

Also an idea that came up during this process was to make the icon buttons black then fade to grey when they appear to draw attention to them.  Could make an issue for backburner if this is a good idea.

### DIFF
--- a/src/components/Requirement.vue
+++ b/src/components/Requirement.vue
@@ -8,34 +8,36 @@
   >
   <v-layout row>
     <v-flex>
-      <div v-if="!leaf">
-        <span v-if="'title-no-degree' in req && req['title-no-degree'] !=''">{{ req["title-no-degree"] }}</span>
-        <span v-else-if = "'short-title' in req && req['short-title'] != ''">{{ req['short-title']}}</span>
-        <span v-else-if = "'title' in req">{{ req["title"] }}</span>
-        <span style="font-style:italic">{{ req['threshold-desc'] }}</span>
-        <!--fake padding for scroll-->
-        &nbsp &nbsp &nbsp
-      </div>
-      <span v-else>
-        <span v-if="'title' in req">{{ req.title }}</span>
-      </span>
-
-      <span v-if = "!req['plain-string']">
-        <span v-if="!('title' in req) && 'req' in req">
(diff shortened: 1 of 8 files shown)


src/components/Requirement.vue
```diff
@@ -3,36 +3,44 @@
...
-      <span v-if = "'title' in req">| </span>
-      <span style = "text-transform: cursive">{{ req.req }}</span>
-    </span>
-    <span v-if = "req.max === 0 && leaf" style = "font-style:italic">
-       (optional)
-    </span>
-    <div :class = "percentage_bar" :style = "percentage"></div>
+  <v-layout row>
+    <v-flex>
+      <div>
```
npfoss: do we need this div? I tested removing it and it didn't seem to break anything, but I assume it's here for a reason?

src/components/Requirement.vue
```diff
@@ -43,8 +51,8 @@ export default {
   data: function() {
     return {
       open: [],
-      viewDialog: false,
-      showDelete: false
+      hoveringOver: false,
+      iconHover: false,
```
npfoss: I've noticed that you started adopting my comma convention recently. It's nice for diffs!

src/components/Requirement.vue
georgiashay: Yeah, I can't remember why that div was there but it's obviously no longer necessary.

npfoss:
(testing our first squash and merge)

PR #153, opened 2019-04-14
About page

### LINKED ISSUE #143: logo and about page
By logo I just mean something that says "Courseroad" in the top left, and opens an about page in a new tab when clicked.

That about page should have at the very least:
- link to github repo
- acknowledgement/thanks to FireRoad and links to those repos
- names of maintainers (us)
- link to old courseroad
- known issues? (with notes about why we're not going to fix some obscure audit oddity etc)
- ?

georgiashay:
Closes #143 

Moved App.vue into pages and renamed to MainPage.vue, so we could have two pages (MainPage and About).  Using VueRouter now to go between them.  This requires App.vue to have a <router-view> where these two pages should be inserted, and <router-links> when you want to switch between the pages.  It also made it necessary to change how we handle the #road[roadID] in the url, since the router inserts a #/.  

Might not be 100% ready for a merge, could potentially use input on:
- UI: logo and about page don't look super-great, but this might not matter for MVP
- Known issues: I put two in there, but I couldn't think of any others.  It seems like most of the weird audit cases are now being handled.


npfoss:
I'm concerned about merging this since it likely breaks things other people are working on because App.vue moved. How should we go about this...?

georgiashay:
Will github understand it's the same file?  Other than moving/renaming, there was just a v-app deleted in that file.

georgiashay:
Nope.  Just tested it and this will definitely be a problem.  I think there may be some way of telling git that you didn't create a new file, you just moved an old one?

ttpcodes:
I'm going to look at this branch tomorrow and see if I can rebase the branch to break apart the commits so that App.vue doesn't cause conflicts due to it being moved.

georgiashay:
This will cause merge conflicts

PR #164, opened 2019-04-17
Placeholder Hours

### LINKED ISSUE #162: Placeholder classes have no hours
This makes the total # of hours shown at the top of the semester artificially lower.

Some options
- Put in the average hours for all matching classes (would also show in the class info card though, which is not necessarily what we want)
- Put in 12 hours (may not be very accurate)
- Put in average hours but don't show in class info card (check if it's a generic course)

I think this should be addressed now since it will likely affect many people and could lead them to make poor decisions.

georgiashay:
Averages hours for placeholders/generic courses.  Displays in class info card with rounding to two decimal places.  Class info card label changes to "average hours" for these generic courses.

Closes #162

### DIFF
--- a/src/components/ClassInfo.vue
+++ b/src/components/ClassInfo.vue
@@ -90,8 +90,8 @@
                   <td><b>Hours</b></td>
                   <td>
                     <table cellspacing = "0">
-                      <tr v-if = "currentSubject.in_class_hours !== undefined">{{currentSubject.in_class_hours}} in class</tr>
-                      <tr v-if = "currentSubject.out_of_class_hours !== undefined">{{currentSubject.out_of_class_hours}} out of class</tr>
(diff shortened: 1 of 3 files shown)


npfoss:
I think noting "average hours" in the card is good. Do you think we need to add something similar in the semester header? I'm actually leaning towards replacing "Hours" with "Estimated Hours" for everything, not just semesters with generics

georgiashay:
Hmm, yeah maybe some sort of disclaimer would be good in the header as well.  Where would the asterisk lead to though?  Or would Est Hours instead of Hours be good enough to cover us?

npfoss:
I think Est Hours is enough

venkatesh-sivaraman:
+1 for using “Estimated Hours“, but also one thing that could subconsciously reinforce this is displaying the number to one decimal place. This contrasts it with the adjacent units number and makes it look more “calculated” and less official.

PR #170, opened 2019-04-17
need to use the right fireroad link

npfoss:
Hey @ttpcodes, could you take over this PR and add `FIREROAD_LINK` (in both `Auth.vue` and `MainPage.vue`) to the webpack stuff like the other thing please? It should be `fireroad-dev` for local and /dev/, and `fireroad` for prod.

### DIFF
--- a/build/webpack.config.dev.js
+++ b/build/webpack.config.dev.js
@@ -43,7 +43,8 @@ module.exports = (env) => {
         template: 'index.html',
         inject: true
       }),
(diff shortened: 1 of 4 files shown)


georgiashay:
I can do this

ttpcodes:
@georgiashay Don't worry about it, I already started working on the PR.

ttpcodes:
Assuming I didn't miss anything, these two commits should set `FIREROAD_URL` from `webpack` now.

npfoss:
I tested as much as I could and it looks good!

PR #171, opened 2019-04-17
Redirect mobile users to FireRoad

### LINKED ISSUE #145: mobile support
As in, we should link them to FireRoad if they're on mobile.

Suggestion and useful links from Venkat:
> Before I set you up on the prod server, could you do one thing for me? The site is pretty hard to use on mobile, and that’s fine because there’s FireRoad for that. So I think it would be better if on iOS/Android browsers, instead of showing the UI at all you just prompt the user to download FireRoad: something like “Access your roads and more with FireRoad on mobile.” You can use the icon at https://fireroad.mit.edu/catalogs/fireroad_logo.png. And the link to the appropriate app store is http://onelink.to/stk7us.

ttpcodes:
This PR adds `User Agent` detection via `ua-parser-js` in order to determine if a user is on a mobile browser or not. If they are a mobile browser, the app will prompt them to download `FireRoad` for their respective platform as an alternative, with the option to continue using the website if they really wish to.

This PR closes #145.

### DIFF
--- a/src/pages/MainPage.vue
+++ b/src/pages/MainPage.vue
@@ -2,6 +2,27 @@
 <template>
   <v-app id="app-wrapper"
   >
+    <v-dialog v-model="showMobile" fullscreen>
+      <v-card height="100%">
+        <v-container fill-height>
+          <v-layout column>
+            <v-flex grow>
+              <v-card-title primary-title>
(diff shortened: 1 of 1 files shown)


georgiashay:
Pushed to dev; verified it works

PR #172, opened 2019-04-17
Toolbar icons

georgiashay:
Made toolbar have login/logout, import/export icons when screen is small
Made the class search resize to fit on small screens
Added scrollbar if the class search can't fit the filters (I know we have disliked scroll bars in the past, so if this looks terrible we should remove it, it's the last commit)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,6 +2,4 @@ node_modules
 dist
 .idea
 .babelrc
-.package-lock.json
-package-lock.json
 *~
--- a/build/webpack.config.dev.js
+++ b/build/webpack.config.dev.js
@@ -43,7 +43,8 @@ module.exports = (env) => {
         template: 'index.html',
         inject: true
       }),
-      new webpack.DefinePlugin({ 'process.env.APP_URL': JSON.stringify(env.APP_URL) })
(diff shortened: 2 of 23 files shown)


npfoss:
I'm too tired, and this is noncritical for MVP, sorry... I do appreciate it though and will take a look later.

georgiashay:
don't merge this yet something is wrong with it

georgiashay:
image

georgiashay:
This should be ready to merge now

wishdasher:
The search bar is no longer set with the right side of the screen. Is this intentional, and if so, why?

georgiashay:
Not intentional - thought it was at the right side last night but I’ll take a look later today

npfoss:
oh, dammit, sorry I didn't catch that. thanks for noticing.

I just tried the simple solution of adding a `<v-spacer>` and that wasn't sufficient fyi

georgiashay:
This is what it looks like for me

image

wishdasher:
image

wishdasher:
I think you have to zoom out a bit.

georgiashay:
I think this is because of the resizing of said searchbar on small screens that I added.  Will look into this

npfoss:
This is now fixed, for the record

PR #177, opened 2019-04-18
A bunch of small styling changes

wishdasher:


src/components/ImportExport.vue
```diff
@@ -19,6 +19,8 @@
             outline
             label="Road name"
             clearable
+            v-if="dialog"
```
npfoss: why do we need this?

src/components/Auth.vue
```diff
@@ -1,38 +1,44 @@
...
 </template>
 
 <script>
-import Vue from 'vue'
-import UAParser from "ua-parser-js"
-var DATE_FORMAT = "YYYY-MM-DDTHH:mm:ss.SSS000Z"
+import Vue from 'vue';
+import UAParser from "ua-parser-js";
+var DATE_FORMAT="YYYY-MM-DDTHH:mm:ss.SSS000Z";
+var FIREROAD_LINK="https://fireroad-dev.mit.edu";
```
npfoss: I think this line got caught in a merge from somewhere else, it's no longer needed.

src/components/Requirement.vue
```diff
@@ -30,10 +36,29 @@
...
+          style="padding-left: 0.2em; padding-right: 0em;"
+          v-if="'reqs' in req && hoveringOver"
+          @mouseover = "iconHover = true"
+          @mouseleave = "iconHover = false"
+          @click.stop = "$emit('click-info', $event)"
+          small
+          :color = "iconColor"
+        >
+          info
+        </v-icon>
```
npfoss: I appreciate that this is above the bar now so it doesn't change size when you hover

src/components/ImportExport.vue
npfoss: oh, is it so the autofocus behaves properly?

src/pages/MainPage.vue
```diff
@@ -87,6 +87,7 @@
           v-model = "searchInput"
           placeholder = "Add classes"
           slot = "activator"
+          autofocus
```
npfoss: I just had the idea to add this to draw attention to the way you add classes!

src/components/Semester.vue
```diff
@@ -1,19 +1,27 @@
...
-          </b>
-          Units: {{semesterInformation.totalUnits}}
-          Est. Hours: {{semesterInformation.expectedHours.toFixed(1)}}
+          <span style="width: 6em; display: inline-block;">
+            <b>
+              {{semesterType}}
+              <v-hover>
+                <span slot-scope = "{ hover }" @click = "changeYear" :class = "hover && 'hovering'">
+                  {{semesterYear}}
+                </span>
```
npfoss: This is much better than the green highlight

src/components/Auth.vue
```diff
@@ -1,38 +1,44 @@
 <template>
-  <v-layout row align-baseline>
-    <v-btn v-if = "!loggedIn" outline round color = "primary" @click="loginUser">
+  <v-layout row align-center>
+    <v-btn v-if="!loggedIn" outline round color="primary" @click="loginUser">
```
npfoss: Why remove all the spaces around `=`s? Honestly I have no idea what standard linters do but I always thought space around assignment `=` signs was easier to read.

src/components/ImportExport.vue
wishdasher: [Yeah](https://stackoverflow.com/questions/51472947/vuetifys-autofocus-works-only-on-first-modal-open)

src/components/Auth.vue
wishdasher: I'm not sure what standard linters do either, but the examples [here](https://vuejs.org/v2/style-guide/) don't use the spaces. A close parallel I'm familiar with is props for React components, which don't use the spaces either (`<Blah prop={x} />`).

I personally removed them because my code editor doesn't do syntax highlighting correctly for the stuff inside the `""` unless those spaces aren't there.

wishdasher:
Your fixes look good too! I will continue to make small tweaks but probably in a separate PR.

src/components/Auth.vue
wishdasher: Ah, I agree that normal assignment operators shouldn't have them though. I'll go through and try to reverse those cases.

npfoss: Oh, yeah, I left this comment in a bad place. I totally agree for things like this, because they're like keyword arguments. Should have chosen an assignment line

npfoss: And also a lot of conditionals have `a===b`, which I think is nonstandard

wishdasher: lgtm (commit 659189e)

npfoss: oh, no, there are more, I thought it would be a quick regex match thing but there were a lot that didn't pick up and I'm trying to go to bed :P

wishdasher: oops :disappointed_relieved:

npfoss: it looks like a lot of it happened in this commit: b6d9fe5ebcad07c60ef92fe778d2ef9f93e18838

so you might be able to just rebase it out. there were some other good changes in there though so it might be a pain either way

src/components/Auth.vue
```diff
@@ -375,29 +368,30 @@ export default {
...
               tabs.push(this.tabID);
               this.$cookies.set("tabs", JSON.stringify(tabs));
             }
           } else {
             this.$cookies.set("tabs", JSON.stringify([this.tabID]));
           }
         } else {
-          if(this.$cookies.isKey("tabs") && (tabs=JSON.parse(this.$cookies.get("tabs"))).length) {
+          // TODO: look into whether this = sign is acting correctly?
+          if (this.$cookies.isKey("tabs") && (tabs = JSON.parse(this.$cookies.get("tabs"))).length) {
```
npfoss: @georgiashay

PR #189, opened 2019-04-19
Fix search results padding

### LINKED ISSUE #187: search description indent
(not urgent)

this weird thing is back:
image

wishdasher:
fixes #187 
image

### DIFF
--- a/src/components/ClassSearch.vue
+++ b/src/components/ClassSearch.vue
@@ -25,15 +25,15 @@
                 slot-scope="{ hover }"
                 :class="{ 'elevation-3': hover }"
                 draggable="true"
-                style="cursor: grab; margin: 4px; display: block;"
+                style="cursor: grab; margin: 4px;"
                 @dragstart="dragStart($event, props)"
                 @click="viewClassInfo(props)"
               >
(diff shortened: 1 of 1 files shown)


wishdasher:
Also slightly condensed the view.

wishdasher:
Sorry, I didn't do a full comparison of the before and after thing (was just trying to fix the alignment). I'll take another look at this and see if I can preserve whatever shading was there before.

wishdasher:
I intend to add the popout back.

wishdasher:
Okay, I spent too much time on this.
Adding reference links to the issue if anyone is interested.
[why the original spacing breaks](https://stackoverflow.com/questions/22392433/auto-adjust-column-widths-of-displayblock-table-rows)
[adding margin to `tr`](https://stackoverflow.com/questions/10690299/how-to-add-a-margin-to-a-table-row-tr)
[random advice](https://stackoverflow.com/questions/5670402/why-is-my-html-table-not-respecting-my-css-column-width)
Googling something like margins on `tr` may be useful.

npfoss:
Oh, well, ok then. I guess "it's impossible" is why we were having so much trouble with this. Since Michael hasn't said otherwise I'm going to merge this because I think it looks better than before despite the loss of left shadows.

PR #190, opened 2019-04-19
Show fewer lines for class cards on wider screens

wishdasher:
When the screen is wider, class titles generally fit onto the cards.
image
image


wishdasher:
The cards may look small, but the vertical real estate is nice.

npfoss:
It looks pretty good in your screenshot, but it make my computer look like this by default...
image

It's probably better to err on the side of more room since people can always collapse past semesters

wishdasher:
This was intentional, for the full-width screen to generally only show two lines because most class titles aren't so long. I feel like more vertical space is useful, especially when you're dragging around things to plan lots of future semester as a freshman.

npfoss:
Maybe the solution is that this becomes our first adjustable setting ("compact mode"), because I'm hesitant to make the default something that renders `/Introduction to .*/` and `/Fundamentals of .*/` unreadable, especially if the intended beneficiaries are freshmen:
image

We wanted to add other settings eventually anyways, like dark mode, etc

wishdasher:
I like the settings idea. Also, on Firefox and some other browsers, the long words should have auto-hyphentation. For some reason, Chrome is behind on this feature,

georgiashay:
Maybe it would be better if the x button didn't take up the whole left hand side, and words could wrap underneath it as well?

georgiashay:
Do you think it would be possible to programmatically figure out a shorter title to display?  Or at least remove Introduction and other generic words?

venkatesh-sivaraman:
That's a really slick idea, and probably could be done with a simple list of replacements. I'd definitely opt for "Fun of Programming" on my road over "Fundamentals of Programming", lol. I still think the title sizes could in general be smaller, regardless of tile size - right now the title font is bigger than the semester headings.

If you'd like to actually make the settings menu and have a "compact mode", I wonder if you could have three different sizes like Gmail does with the default/comfortable/compact modes. That could change the tile size, text size, and spacing between and within tiles.

npfoss:
Those all seem like good ideas to me. It also might just be better to reduce the padding on both the headers and the body of the `v-expansion-panels` because we don't lose information that way. I just spent a few minutes trying to do this and it would look good if I could target the right things but unfortunately the only way I've had success with so far is `.container.lighten-3.fluid.grid-list-md.grey`...

PR #198, opened 2019-04-23
Open search on click

georgiashay:
Not sure most of the props are actually being used correctly (input-activator is supposed to be for v-inputs, not v-text-fields), but it only looks/behaves right with them like this.

Conflict on close-on-click, unexpected behavior both ways
* If true, then you can close by clicking, then keep typing and it won't open until you click again
* If false, then you can't close by clicking (which I think is more intuitive)

Could solve either of these by going back to the v-model of showSearch, but this seems to be what was preventing it from working properly in the first place.


georgiashay:
Thought that functionality wasn't removed, I'll see what happened there

georgiashay:
Ok, tried to wrestle with the props again but they're just not meant for this.  Moved the text field outside of the menu, which means it has to be positioned using jquery ( :( ).  But it also means that we don't have to use the activator slot and rely on vuetify's definition of "has been activated" and instead added functions to open the menu explicitly when the user types or when it is clicked. 

Lots of merge conflicts since this comes from before the toolbar stuff and the linter, but I'll fix it

georgiashay:
Lots of errors in the linter but I'm not sure which ones are from my PR - how are we handling this currently?

ttpcodes:
There's a number of `linter` PRs that are still open, so I don't think we have a flow even ready to handle this. I guess the best we can do right now is to make sure that between `master` and this PR, that the number of errors hasn't somehow gone *up*.

georgiashay:
Hmm I think maybe the number of errors went up by like 20ish but I couldn't tell what they were from.

npfoss:
I think the solution may be to merge a bunch of the outstanding things and start using Travis for a baseline and make sure Travis-checked PRs never increase the error count

PR #200, opened 2019-04-23
Manually fix lint warnings and use `no-var` where possible

wishdasher:
Now with only 117 problems (46 errors, 71 warnings)!


src/components/ClassSearch.vue
```diff
@@ -16,7 +16,7 @@
           :rows-per-page-items="rowsPerPageItems"
           :pagination.sync="pagination"
           :no-data-text="'No results'"
-          :rows-per-page-text="'Results per page: '"
+          :rows-per-page-text="'Results per page:\xa0'"
```
ttpcodes: What happened here?

src/pages/MainPage.vue
```diff
@@ -36,7 +36,7 @@
         :roads="roads"
         :active-road="activeRoad"
         :subjects="subjectsInfo"
-        @delete-road="$refs.authcomponent.deleteRoad($event)"
+        @devare-road="$refs.authcomponent.devareRoad($event)"
```
ttpcodes: Is this supposed to be renamed?

src/pages/MainPage.vue
```diff
@@ -58,7 +58,7 @@
         :just-loaded="justLoaded"
         :active-road="activeRoad"
         :conflict-info="conflictInfo"
-        @delete-road="deleteRoad"
+        @devare-road="devareRoad"
```
ttpcodes: Is this supposed to be renamed?

src/pages/MainPage.vue
```diff
@@ -85,7 +85,7 @@
             id="searchInputTF"
             slot="activator"
             v-model="searchInput"
-            autocomplete="false"
+            autocompvare="false"
```
ttpcodes: Is this supposed to be renamed?

src/pages/MainPage.vue
```diff
@@ -453,10 +453,10 @@ export default {
       if (this.activeRoad === oldid) {
         this.activeRoad = newid;
       }
-      Vue.delete(this.roads, oldid);
+      Vue.devare(this.roads, oldid);
```
ttpcodes: Is this supposed to be renamed?

src/pages/MainPage.vue
```diff
@@ -521,8 +521,8 @@ export default {
     changeActiveRoad: function (event) {
       this.activeRoad = event;
     },
-    deleteRoad: function (roadID) {
-      Vue.delete(this.roads, roadID);
+    devareRoad: function (roadID) {
```
ttpcodes: There's a lot of this `delete` being replaced by `devare`. That probably should be fixed.

src/components/ClassSearch.vue
wishdasher: Originally, I copy+pasted a non-standard whitespace character. `\xa0` is a non-breaking space, which is a better way of doing this.

src/pages/MainPage.vue
wishdasher: Nope, result of bad find+replace. Thanks for catching it!

src/components/ClassSearch.vue
wishdasher: More context: `eslint` didn't like the string having extra whitespace at the end.

npfoss:
I'm getting a ton of console errors just from logging in and loading roads:
image
but I think I have figured out the problem and will push the solution shortly

PR #201, opened 2019-04-23
Manual code style changes

wishdasher:
Mostly ternaries, basing this off branch `lint2` (PR #200) for now.

### DIFF
--- a/src/components/ClassInfo.vue
+++ b/src/components/ClassInfo.vue
@@ -211,50 +211,35 @@ export default {
   computed: {
     currentSubject: function () {
       const currentID = this.classInfoStack[this.classInfoStack.length - 1];
-      let curSubj;
-      if (currentID in this.subjectsIndex) {
-        curSubj = this.subjects[this.subjectsIndex[currentID]];
-      } else {
-        curSubj = this.genericCourses[this.genericIndex[currentID]];
-      }
-      return curSubj;
+      return currentID in this.subjectsIndex
+        ? this.subjects[this.subjectsIndex[currentID]]
+        : this.genericCourses[this.genericIndex[currentID]];
     },
     parsedPrereqs: function () {
-      if (this.currentSubject.prerequisites !== undefined) {
-        return this.parseRequirements(this.currentSubject.prerequisites);
-      } else {
-        return {
-          reqs: []
-        };
-      }
+      return this.currentSubject.prerequisites !== undefined
+        ? this.parseRequirements(this.currentSubject.prerequisites)
+        : { reqs: [] };
     },
     parsedCoreqs: function () {
-      if (this.currentSubject.corequisites !== undefined) {
-        return this.parseRequirements(this.currentSubject.corequisites);
-      } else {
-        return {
-          reqs: []
-        };
-      }
+      return this.currentSubject.corequisites !== undefined
+        ? this.parseRequirements(this.currentSubject.corequisites)
+        : { reqs: [] };
     }
   },
   methods: {
     classInfo: function (subjectID) {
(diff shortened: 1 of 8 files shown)


wishdasher:
Should subject this to a little more rigorous reviewing

georgiashay:
Some of these optimizations are definitely better but I’m not sure we want to change all of them to ternaries.  I think if statements are more readable sometimes but I’m not sure which is best in some of these cases so I would like some second opinions here

Also is !! always safe, and is it readable enough to use?

ttpcodes:
Code looks good to me. With this being a code styling issue though, I'd like others to voice what their opinions are. It seems to me though that these situations (trying to choose between strings) are good places to use ternary operators, but that's just my opinion.

As for double exclamation marks to cast to `boolean`, I'm a bit more conflicted. I've actually never really had to cast **anything** to `boolean` in JavaScript, but I have seen this used in some projects that I now maintain (like Next House's Resident Dashboard, or `nextres`). It seems to be a safe way to cast to `boolean`, but the general complaint is that it's not intuitive to read. I've heard various cursed things about the alternative though, which is using the `Boolean` constructor, so I'm not sure if there really is any other way to cast to `boolean` that makes sense.

Once we come to a consensus on this, I'll be willing to approve it, but I'm not quite comfortable approving this just yet with the other details still up in the air.

src/components/ClassInfo.vue
```diff
@@ -211,50 +211,35 @@ export default {
...
-      } else {
-        return {
-          subject_id: subjectID,
-          title: ''
-        };
-      }
+      return subj || {
+        subject_id: subjectID,
+        title: ''
+      };
```
npfoss: ooh, fancy. I like this, but is it exactly equivalent to the undefined thing or is it actually casting to a boolean or something?

src/components/ClassSearch.vue
```diff
@@ -164,6 +164,7 @@ export default {
           'level': getRegexFuncs(this.chosenFilters.levelInput),
           'total_units': getRegexFuncs(this.chosenFilters.unitInput)
         };
+        // TODO: remove?
```
npfoss: @georgiashay I think you kept this commented out stuff around for a reason earlier, but that's probably unnecessary now right?

georgiashay: Left it for discovering all values of a given attribute on a subject.  Could still be useful but it'd be better to just copy paste somewhere and use it when needed than keep it commented out in the codebase.

src/components/Semester.vue
```diff
@@ -267,9 +261,9 @@ export default {
         let inPreviousQuarter = false;
         if (subj2 !== undefined) {
           inPreviousQuarter = s.semester === this.index &&
-                                  subjInQuarter2 &&
-                                  subj2.quarter_information !== undefined &&
-                                  subj2.quarter_information.split(',')[0] === '0';
+            subjInQuarter2 &&
+            subj2.quarter_information !== undefined &&
+            subj2.quarter_information.split(',')[0] === '0';
```
npfoss: I am personally a fan of putting the `&&`s at the beginning of the line and think that's a lot more readable, but if the linter says this is better then maybe it is

src/components/ClassInfo.vue
ttpcodes: This is a short-circuit `OR` evaluation: it returns the first item that is "truthy". This means that is `subj` is undefined, it'll return the object. Otherwise, it'll return `subj`.

src/components/Semester.vue
ttpcodes: Personal preference, but I like keeping the `&&` and other operators on the same line.

PR #203, opened 2019-04-24
Remove all CDN resources

### LINKED ISSUE #194: Load axios and moment locally
Right now, `axios` and `moment` are loaded from `CDN`, which we should try to avoid as seen by what happened with `jsdelivr` earlier.

ttpcodes:
This PR removes the remainder of `CDN` resources that were referenced in `index.html` and replaces them with our own bundles, to avoid issues with a `CDN` going down again. Additionally, I've changed the `webpack` configuration a bit, such that it separates `CSS` into its own file so that this new bundle works properly and so that it's easier to debug `webpack` issues in the future.

This PR closes #194.


build/webpack.config.dev.js
```diff
@@ -16,6 +17,17 @@ module.exports = (env) => {
     },
     module: {
       rules: [
+        {
+          test: /\.(woff(2)?|ttf|eot|svg)(\?v=\d+\.\d+\.\d+)?$/,
```
georgiashay: Not entirely sure what is going on here.  Seems like this part loads fonts but the regex is not clear.  Could use a comment to explain.

ttpcodes: Yep, this is related to font parsing. It does some weird things to parse various font files and other extensions that might be appended to the end of the file. I remember getting this off of some `webpack ` documentation, but I don't remember exactly where. It ended up on another one of my `webpack` projects though, and I took it from there and put it here.

ttpcodes:
The regex should be commented on now.

PR #206, opened 2019-04-24
Justify searchbar

georgiashay:
Justifies the searchbar to the right again
Not 100% sure if it's bad to have both scoped and non-scoped styles but wasn't able to figure out the css on this without directly modifying the classes vuetify produces.  This is probably bad anyway, but none of the vuetify-supplied things were working for me.

### DIFF
--- a/src/pages/MainPage.vue
+++ b/src/pages/MainPage.vue
@@ -70,7 +70,7 @@
         @set-sem="setSemester"
       />
 
-      <v-layout grow>
+      <v-layout>
         <v-menu
           v-model="showSearch"
           attach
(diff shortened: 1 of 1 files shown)


georgiashay:
I have tried various combinations of v-layouts/v-flex including something similar to this I think, but none of them did exactly what I wanted.

Changed how the menu works in #198, so the changes I made won't work anymore if that goes through.  don't merge this yet, will probably have to redo

georgiashay:
Turns out it is a lot easier to justify when it's not inside a menu.  Vuetify prop justify-end actually works in this case.  Adding to #198.

npfoss:
fixed in #198

PR #223, opened 2019-04-27
fix default year

### LINKED ISSUE #221: Year shown is incorrect
Year starts at Fall 2019 even though it is set to Freshman

npfoss:
fixes #221 
also a temporary fix for #238 

~The reason it's showing fall 2019 now is because the default semester is 0 and the math to figure out when that is works. Unfortunately that means we'll have to use a different cue for whether they've manually set their roads or not, which is probably better anyways (going to the source).~
Makes default semester 1. Rolls year and class-classification over at the end of May instead of end of April. Doesn't mess with the way FireRoad does it by adding/subtracting 1 when saving/getting current year during May only.

I also removed some code that was chopping off your last semester if you change your year.

### DIFF
--- a/src/components/Auth.vue
+++ b/src/components/Auth.vue
@@ -153,10 +153,11 @@ export default {
       const headerList = { headers: {
         'Authorization': 'Bearer ' + this.accessInfo.access_token
       } };
+      const currentMonth = new Date().getMonth();
       return axios.get(process.env.FIREROAD_URL + '/verify/', headerList)
         .then(function (verifyResponse) {
           if (verifyResponse.data.success) {
-            this.$emit('set-sem', verifyResponse.data.current_semester - 1);
+            this.$emit('set-sem', verifyResponse.data.current_semester - (currentMonth === 4 ? 1 : 0));
             return verifyResponse.data;
           } else {
             this.logoutUser();
@@ -489,7 +490,7 @@ export default {
       const sem = currentMonth >= 5 && currentMonth <= 10
         ? 1 + year * 3
         : 3 + year * 3;
-      this.postSecure('/set_semester/', { semester: sem + 1 }).then(function (res) {
+      this.postSecure('/set_semester/', { semester: sem + (currentMonth === 4 ? 1 : 0) }).then(function (res) {
         if (res.status === 200 && res.data.success) {
           this.$emit('set-sem', sem);
         }
--- a/src/components/Auth.vue
+++ b/src/components/Auth.vue
@@ -156,7 +156,7 @@ export default {
       return axios.get(process.env.FIREROAD_URL + '/verify/', headerList)
         .then(function (verifyResponse) {
           if (verifyResponse.data.success) {
-            this.$emit('set-sem', verifyResponse.data.current_semester);
+            this.$emit('set-sem', verifyResponse.data.current_semester - 1);
             return verifyResponse.data;
           } else {
             this.logoutUser();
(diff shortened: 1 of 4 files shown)


georgiashay:
If I change my semester back to 0 and test this again I see the same thing.  It shows up as starting in Fall 2019 and when I go to change my year it is already set to Freshman (which would lead to confusion because it looks like that's already set correctly but the year is wrong). 

The default only matters for not-logged-in-users, because it's going to get the correct semester from fireroad as soon as it can and override the default.  The "correct semester" is initially 0 so the problem happens again. 

In general we shouldn't be setting it to a semester that doesn't make sense based on the calendar.  Right now, only 3,6,9... make sense but pretty soon that will change to 1,4,7...  I think it will mess up the year calculations if we don't do this.  We could set it to automatically to Freshman year in the correct semester by checking the date instead of setting it to 1 always, whenever Fireroad returns a 0, making sure we also send this correct semester back to Fireroad.  

I think maybe a pre-frosh setting would also be valuable?  With current semester = 0?  There's a chance pre-frosh are planning their courses already but probably low enough to not worry about now unless it would help fix this problem.  Having the default be pre-frosh is probably still bad though, since a majority of users are not pre-frosh and this could be confusing.

npfoss:
Ok new stuff:
Makes default semester 1. Rolls year and class-classification over at the end of May instead of end of April. Doesn't mess with the way FireRoad does it by adding/subtracting 1 when saving/getting current year during May only. Also underlines everything now, not just year number.

src/components/Auth.vue
```diff
@@ -153,10 +153,11 @@ export default {
       const headerList = { headers: {
         'Authorization': 'Bearer ' + this.accessInfo.access_token
       } };
+      const currentMonth = new Date().getMonth();
       return axios.get(process.env.FIREROAD_URL + '/verify/', headerList)
         .then(function (verifyResponse) {
           if (verifyResponse.data.success) {
-            this.$emit('set-sem', verifyResponse.data.current_semester);
+            this.$emit('set-sem', verifyResponse.data.current_semester - (currentMonth === 4 ? 1 : 0));
```
georgiashay: This same logic should also apply to line 120, `this.$emit('set-sem', this.accessInfo.current_semester);`

src/components/Auth.vue
```diff
@@ -486,10 +487,10 @@ export default {
     },
     changeSemester: function (year) {
       const currentMonth = new Date().getMonth();
-      const sem = currentMonth >= 4 && currentMonth <= 10
+      const sem = currentMonth >= 5 && currentMonth <= 10
```
georgiashay: I think this is fine for the quick fix but we should also apply this same logic to the other bound; right now in November it will switch to Spring (or maybe IAP I forget).

src/pages/MainPage.vue
```diff
@@ -320,17 +320,14 @@ export default {
       searchInput: '',
       showSearch: false,
       classInfoStack: [],
-      currentSemester: 0,
+      currentSemester: 1,
```
georgiashay: This is definitely good for non-logged in users, but FireRoad still has default of 0, so unless we change incoming 0's into 1's, users who have never changed their year will still have Fall 2019 show up as the first year.  If we want to implement this it would go in the verify function after you receive the current semester on the FireRoad server.

src/components/Auth.vue
npfoss: Good catch, thank you. I forgot the cookie was directly storing what FireRoad gave us

src/pages/MainPage.vue
npfoss: Good call. I put it inside `setSemester` to be safe

src/components/Auth.vue
npfoss: I think that's fine because class year doesn't change then, right? So now it's a question of what's what for highlighting purposes, and I don't think that's important enough to warrant a manual adjustment every time this number goes in or out of FireRoad like we have to for the May switch

src/components/Auth.vue
npfoss: Actually it turns out that line was unnecessary because `verify` is called right below it and that's where this happens.

georgiashay:
If the semester is 0 on FireRoad, it will set the local semester to 1 but not update the semester on FireRoad.  Wanted to make sure this is intended behavior before approving.  

Also, should we default to 3 instead of 1 in the Spring?

npfoss:
decided in person this is fine

PR #226, opened 2019-04-29
Implement Vuex

### LINKED ISSUE #195: Add vuex
Add states/stores/Vuex so we don't have to import genericIndex, subjectsIndex, etc into every thing.

ttpcodes:
This PR adds `vuex` support to `courseroad2` such that we can place commonly used variables in a global context rather than having to pass them down the long `Component` tree like we currently are.

The things that have been moved are as follows:
- `Auth` logic
- `Cookie` logic
- `Road` logic
- `ClassStack` logic

There are a **lot** of things that were moved around, and I don't doubt the possibility that something may have gotten lost in all of these changes and possibly broken something. I've done some preliminary testing on my own, but please also test the changes to make sure I didn't break anything. If any red appears in the `Inspect Element` console, chances are it's because something in my series of changes broke it.

This PR resolves #195.

### DIFF
--- a/build/webpack.config.dev.js
+++ b/build/webpack.config.dev.js
@@ -66,7 +66,8 @@ module.exports = (env) => {
         filename: 'css/app.css'
       }),
       new webpack.DefinePlugin({ 'process.env.APP_URL': JSON.stringify(env.APP_URL) }),
-      new webpack.DefinePlugin({ 'process.env.FIREROAD_URL': JSON.stringify(env.FIREROAD_URL) })
+      new webpack.DefinePlugin({ 'process.env.FIREROAD_URL': JSON.stringify(env.FIREROAD_URL) }),
+      new webpack.DefinePlugin({ 'process.env.NODE_ENV': JSON.stringify(process.env.NODE_ENV) })
     ]
   }
 }
--- a/src/components/Audit.vue
+++ b/src/components/Audit.vue
@@ -53,10 +53,6 @@
         <requirement
           :req="item"
           :leaf="leaf"
-          :subjects="subjects"
-          :subject-index="subjectIndex"
-          :generic-courses="genericCourses"
-          :generic-index="genericIndex"
           @drag-start-class="$emit('drag-start-class',$event)"
           @click.native="clickRequirement(item)"
           @click-info="reqInfo($event, item)"
@@ -162,10 +158,6 @@ export default {
     'selectedReqs',
     'reqTrees',
     'reqList',
-    'subjects',
-    'genericCourses',
-    'subjectIndex',
-    'genericIndex',
     'progressOverrides'
   ],
   data: function () {
--- a/src/components/Audit.vue
+++ b/src/components/Audit.vue
@@ -252,7 +252,7 @@ export default {
     },
     deleteReq: function (req) {
       const reqName = req['list-id'].substring(0, req['list-id'].indexOf('.reql'));
-      this.store.commit('removeReq', reqName);
+      this.$store.commit('removeReq', reqName);
     },
     clickRequirement: function (item) {
       if (item.req !== undefined) {
@@ -261,7 +261,7 @@ export default {
           if (usedReq.indexOf('GIR:') === 0) {
             usedReq = usedReq.substring(4);
           }
-          this.$emit('push-stack', usedReq);
+          this.$store.commit('pushClassStack', usedReq);
         } else {
           this.startProgressDialog(item);
         }
--- a/src/components/Audit.vue
+++ b/src/components/Audit.vue
@@ -180,10 +180,10 @@ export default {
         const currentReqs = this.selectedReqs;
         if (currentReqs.length > newReqs.length) {
           const diff = currentReqs.find(x => !newReqs.includes(x));
-          this.$emit('remove-req', diff);
+          this.$store.commit('removeReq', diff);
         } else {
           const newReq = newReqs[newReqs.length - 1];
-          this.$emit('add-req', newReq);
+          this.$store.commit('addReq', newReq);
         }
       }
     },
@@ -252,7 +252,7 @@ export default {
     },
     deleteReq: function (req) {
       const reqName = req['list-id'].substring(0, req['list-id'].indexOf('.reql'));
-      this.$emit('remove-req', reqName);
+      this.store.commit('removeReq', reqName);
     },
     clickRequirement: function (item) {
       if (item.req !== undefined) {
@@ -306,7 +306,7 @@ export default {
     },
     updateManualProgress: function () {
       if (this.progressReq['list-id'] !== undefined) {
-        this.$emit('update-progress', {
+        this.$store.commit('updateProgress', {
           listID: this.progressReq['list-id'],
           progress: this.newManualProgress
         });
--- a/src/components/Audit.vue
+++ b/src/components/Audit.vue
@@ -53,7 +53,6 @@
         <requirement
           :req="item"
           :leaf="leaf"
-          @drag-start-class="$emit('drag-start-class',$event)"
           @click.native="clickRequirement(item)"
           @click-info="reqInfo($event, item)"
         />
--- a/src/components/Auth.vue
+++ b/src/components/Auth.vue
@@ -183,7 +183,7 @@ export default {
         } };
         return this.verify()
           .then(function (verifyResponse) {
-            return !!params
+            return params
               ? axiosFunc(process.env.FIREROAD_URL + link, params, headerList)
               : axiosFunc(process.env.FIREROAD_URL + link, headerList);
           });
@@ -344,7 +344,6 @@ export default {
                 const conflictInfo = { id: this.oldid, other_name: response.data.other_name, other_agent: response.data.other_agent, other_date: response.data.other_date, other_contents: response.data.other_contents, this_agent: response.data.this_agent, this_date: response.data.this_date };
                 this.data.$emit('conflict', conflictInfo);
               } else {
-
                 this.$store.commit('setRoadProp', {
                   id: newid,
                   prop: 'downloaded',
@@ -364,7 +363,6 @@ export default {
                 } else {
                   return Promise.resolve({ oldid: this.oldid, newid: this.oldid, state: 'same' });
                 }
-
               }
             }
           }.bind({ oldid: roadID, data: this }));
--- a/src/components/Auth.vue
+++ b/src/components/Auth.vue
@@ -431,11 +431,11 @@ export default {
     },
 
     updateLocal: function (roadID) {
-      Vue.set(this.roads[roadID], 'name', this.conflictInfo.other_name);
-      Vue.set(this.roads[roadID], 'agent', this.conflictInfo.other_agent);
-      Vue.set(this.roads[roadID], 'changed_date', this.conflictInfo.other_date);
-      Vue.set(this.roads[roadID], 'contents', this.conflictInfo.other_contents);
-      Vue.set(this.roads[roadID], 'downloaded', moment().format(DATE_FORMAT));
+      this.$store.commit('setRoadProp', {id: roadID, prop: 'name', value: this.conflictInfo.other_name, ignoreSet: false});
+      this.$store.commit('setRoadProp', {id: roadID, prop: 'agent', value: this.conflictInfo.other_agent, ignoreSet: false});
+      this.$store.commit('setRoadProp', {id: roadID, prop: 'changed_date', value: this.conflictInfo.other_date, ignoreSet: false});
+      this.$store.commit('setRoadProp', {id: roadID, prop: 'contents', value: this.conflictInfo.other_contents, ignoreSet: false});
+      this.$store.commit('setRoadProp', {id: roadID, prop: 'downloaded', value: moment().format(DATE_FORMAT), ignoreSet: false});
       this.$emit('resolve-conflict');
     },
 
--- a/src/components/Auth.vue
+++ b/src/components/Auth.vue
@@ -344,7 +344,7 @@ export default {
                 const conflictInfo = { id: this.oldid, other_name: response.data.other_name, other_agent: response.data.other_agent, other_date: response.data.other_date, other_contents: response.data.other_contents, this_agent: response.data.this_agent, this_date: response.data.this_date };
                 this.data.$emit('conflict', conflictInfo);
               } else {
-                this.$store.commit('setRoadProp', {
+                this.data.$store.commit('setRoadProp', {
                   id: newid,
                   prop: 'downloaded',
                   value: moment().format(DATE_FORMAT),
@@ -357,7 +357,7 @@ export default {
                   // i suspect this is because the three events required were not happening
                   // in the correct order or something
                   if (this.oldid !== response.data.id.toString()) {
-                    this.$store.commit('resetID', this.oldid, response.data.id);
+                    this.data.$store.commit('resetID', this.oldid, response.data.id);
                   }
                   return Promise.resolve({ oldid: this.oldid, newid: response.data.id, state: 'changed' });
                 } else {
--- a/src/components/Auth.vue
+++ b/src/components/Auth.vue
@@ -32,7 +32,7 @@
         {{ saveIcon }}
       </v-icon>
       <div>
-        <p v-for="saveWarning in saveWarnings" :key="saveWarning">
+        <p v-for="saveWarning in saveWarnings" :key="saveWarning.name + saveWarning.error">
           {{ saveWarning.name }}: {{ saveWarning.error }}
         </p>
       </div>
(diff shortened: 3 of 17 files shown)


src/components/Audit.vue
```diff
@@ -260,7 +252,7 @@ export default {
     },
     deleteReq: function (req) {
       const reqName = req['list-id'].substring(0, req['list-id'].indexOf('.reql'));
-      this.$emit('remove-req', reqName);
+      this.store.commit('removeReq', reqName);
```
npfoss: missing `$`?

ttpcodes: Huh, not sure how I didn't notice that.

src/pages/MainPage.vue
```diff
@@ -204,31 +159,23 @@
...
+      v-if="$store.state.classInfoStack.length"
       :adding-from-card="addingFromCard"
-      :generic-courses="genericCourses"
-      :generic-index="genericIndexDict"
-      @pop-stack="popClassStack"
-      @push-stack="pushClassStack"
       @add-class="addFromCard"
       @cancel-add-class="cancelAddFromCard"
-      @close-classinfo="classInfoStack = []"
+      @close-classinfo="$store.commit('clearClassInfoStack')"
```
npfoss: For things like this, should we move them down into the component now that we have Vuex or did you leave it this way for easier readability and since this is a huge PR already? I think that's a good argument but also it seems like other things have been moved, so we should probably do all or none for consistency.

build/webpack.config.dev.js
```diff
@@ -66,7 +66,8 @@ module.exports = (env) => {
         filename: 'css/app.css'
       }),
       new webpack.DefinePlugin({ 'process.env.APP_URL': JSON.stringify(env.APP_URL) }),
-      new webpack.DefinePlugin({ 'process.env.FIREROAD_URL': JSON.stringify(env.FIREROAD_URL) })
+      new webpack.DefinePlugin({ 'process.env.FIREROAD_URL': JSON.stringify(env.FIREROAD_URL) }),
+      new webpack.DefinePlugin({ 'process.env.NODE_ENV': JSON.stringify(process.env.NODE_ENV) })
```
npfoss: What's this for?

src/components/Auth.vue
```diff
@@ -279,7 +295,7 @@ export default {
       axios.get(process.env.FIREROAD_URL + '/fetch_token/?code=' + code)
         .then(function (response) {
           if (response.data.success) {
-            if (this.data.authCookiesAllowed) {
+            if (this.data.cookiesAllowed) {
               this.data.$cookies.set('accessInfo', response.data.access_info);
```
npfoss: @georgiashay is `this.data.$cookies` right? It's weird to me that there's a `$` in there

georgiashay: Yeah, the $cookies is correct.  It's how the vue-cookies library works.

georgiashay:
> Clicking on a class and having the info card come up seems to be broken, as is removing a class by clicking the X.

I'm not seeing the problem with the class info card, but I have noticed that removing a class by clicking the X is _really, really_ slow.  (Same for adding classes).  This seems proportional to how many classes there are, I don't notice a difference.

src/stores/courseData.js
```diff
@@ -0,0 +1,232 @@
...
+Vue.use(Vuex);
+
+const DATE_FORMAT = 'YYYY-MM-DDTHH:mm:ss.SSS000Z';
+
+const store = new Vuex.Store({
+  strict: process.env.NODE_ENV !== 'production',
+  state: {
+    activeRoad: '$defaultroad$',
+    classInfoStack: [],
+    cookiesAllowed: false,
```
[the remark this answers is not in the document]
georgiashay: Yes, it should start as undefined.

src/stores/courseData.js
```diff
@@ -0,0 +1,232 @@
...
+        }
+      }
+    },
+    subjectsIndex: {},
+    subjectsInfo: []
+  },
+  mutations: {
+    addClass (state, newClass) {
+      state.roads[state.activeRoad].contents.selectedSubjects.push(newClass);
+      Vue.set(state.roads[state.activeRoad], 'changed', moment().format(DATE_FORMAT));
```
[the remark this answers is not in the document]
georgiashay: Do you mean you deleted the line entirely, or deleted the Vue.set part?  I think we can safely get rid of the Vue.set, but the changed property definitely needs to be set for it to work correctly with fireroad and sync to the server.

src/pages/MainPage.vue
```diff
@@ -423,74 +365,45 @@ export default {
...
-        this.genericIndexDict = this.genericCourses.reduce(function (obj, item, index) {
-          obj[item.subject_id] = index;
-          return obj;
-        }, {});
-      });
+    this.$store.dispatch('loadAllSubjects').then(() => {
+      console.log('Subjects were loaded successfully!');
+    }).catch((e) => {
+      console.log('There was an error loading subjects: \n' + e);
+    });
```
georgiashay: Is this for debugging or should we display this in console on prod?

ttpcodes: I think it would be nice to display errors that occur in the production console (provided we ensure we don't accidentally leak sensitive credentials). That way, we can just tell users to provide us with a log that might make debugging on our end easier.

build/webpack.config.dev.js
ttpcodes: Ah, I needed a way to pass in the current build environment to properly initialize Vuex. Essentially, if this is built for `development`, then it'll enable `strict` mode, while if this is built for `production`, it'll disable it.

This `strict` mode is unrelated to `ECMAScript`'s `strict` mode. `Vuex` `strict` mode means that it throws errors anytime we modify a `store` value without properly calling either a `mutation` or an `action`. This allows us to better debug the code because all of our modifications to these values can be traced to just those `functions` rather than to some random place where we decided to assign the value manually.

We need to disable `strict` mode in `production` since while it's fine for `development`, it adds a lot of `watchers` that weighs down the speed of the application in `production`.

src/stores/courseData.js
```diff
@@ -0,0 +1,232 @@
...
+        'CI-HW': ['Communication Intensive with Writing', 'hw']
+      };
+      const genericCourses = [];
+      const baseGeneric = {
+        description: 'Use this generic subject to indicate that you are fulfilling a requirement, but do not yet have a specific subject selected.',
+        total_units: 12
+      };
+      // biol:b, chem: c, lab: l1, partial lab: l2, rest: r, calc1: m1, calc2: m2, phys1: p1, phys2: p2
+      // hass-a: ha, hass-h: hh, hass-s: hs, hass elective: ht, hass subject: h%5Bahst%5D
+      // commun_int - cih: hc, cihw: hw
```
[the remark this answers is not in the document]
ttpcodes: Anyone else have comments on these comments? I want to get a second opinion before inadvertently deleting any important comments.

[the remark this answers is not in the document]
georgiashay: I believe that was just for my reference as I was constructing the three attribute dicts.  The only thing that might be useful to note is the `hass subject: h%5Bahst%5D` part of the comment, since we don't have any generic HASS classes (just HASS-X), and this would be useful information if we wanted to do that in the future, but it shouldn't be left as a comment here.

src/stores/courseData.js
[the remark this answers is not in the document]
npfoss: I didn't mean that but in retrospect you're right. We can probably remove lots of `Vue.set`s then as long as their effects are preserved

src/pages/MainPage.vue
npfoss: We talked about this the other day and decided that having just a few logs show in production would do more good than harm because it could make debugging way easier and if the site is already broken showing some error messages in the console isn't going to make things worse.

npfoss:
> I'm not seeing the problem with the class info card, but I have noticed that removing a class by clicking the X is _really, really_ slow. (Same for adding classes). This seems proportional to how many classes there are, I don't notice a difference.

I deleted `node_modules`, reinstalled, and tested in both Firefox and Chrome, and wasn't able to delete classes by clicking the X or open the info card by clicking on the ones in the road. Adding a road was slow, but not that slow, and it definitely happened, so I think something else is failing. Maybe events are getting blocked weirdly somewhere? If this is only happening to me I guess I can try to figure out the root cause, but it's definitely a problem.

src/stores/courseData.js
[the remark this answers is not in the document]
georgiashay: Yeah, I think a lot of them can be removed but we have to be careful because some are actually necessary (when manipulating arrays or adding new properties that weren't there before).

georgiashay:
I ran npm install again, and now I am seeing the same problem.  I think I found the source though, looks like the $emits from Class.vue weren't actually transferred over to vuex.

src/pages/MainPage.vue
```diff
@@ -179,22 +143,13 @@
...
-            :subjects="subjectsInfo"
             :road-i-d="roadId"
             :current-semester="currentSemester"
             :adding-from-card="addingFromCard && activeRoad===roadId"
             :item-adding="itemAdding"
-            :subjects-index="subjectsIndexDict"
-            :generic-courses="genericCourses"
-            :generic-index="genericIndexDict"
             :drag-semester-num="(activeRoad===roadId) ? dragSemesterNum : -1"
             @add-at-placeholder="addAtPlaceholder"
```
georgiashay: Moving this function into Class.vue would save a lot of $emits.

src/pages/MainPage.vue
```diff
@@ -179,22 +143,13 @@
...
-            @add-class="addClass"
-            @move-class="moveClass($event.classIndex,$event.semester)"
-            @remove-class="removeClass"
-            @click-class="pushClassStack($event.id)"
             @change-year="$refs.authcomponent.changeSemester($event)"
-            @override-warnings="overrideWarnings($event.override,$event.classInfo)"
             @drag-start-class="dragStartClass"
```
georgiashay: This function also seems to have many emits, so would benefit from being moved into the store. These are optimizations that we may want to deal with after this PR, though.

src/stores/courseData.js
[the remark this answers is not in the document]
ttpcodes: I'l remove the comments and move them to Mattermost or something until we find a better place to put it.

src/stores/courseData.js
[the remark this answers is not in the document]
ttpcodes: Is this something we should do in this PR? I think fixing reactivity might be beyond the scope of this PR and be better left for another PR.

src/pages/MainPage.vue
ttpcodes: That's just something I missed. It definitely should be moved down into the component. There are probably other things that I also missed.

src/pages/MainPage.vue
ttpcodes: Wait, what function are you referring to here?

src/pages/MainPage.vue
ttpcodes: Hmm, now that you mention it, `dragStartClass` is emitted across the board a fair bit. I can probably do the work as part of this PR.

src/pages/MainPage.vue
ttpcodes: Just kidding, I see what you're referring to now (I'm quite tired). Will go ahead and move that.

src/components/ImportExport.vue
```diff
@@ -169,10 +176,10 @@ export default {
...
             let subject;
-            if (this.subjectsIndex[s.id] !== undefined) {
-              subject = this.subjects[this.subjectsIndex[s.id]];
-            } else if (this.genericIndex[s.id] !== undefined) {
-              subject = this.genericCourses[this.genericIndex[s.id]];
+            if (this.$store.state.subjectsIndex[s.id] !== undefined) {
+              subject = this.$store.state.subjectsInfo[this.$store.state.subjectsIndex[s.id]];
+            } else if (this.$store.state.genericIndex[s.id] !== undefined) {
+              subject = this.$store.state.genericCourses[this.$store.state.genericIndex[s.id]];
             }
```
georgiashay: Line 203: `this.$emit('add-road', this.roadtitle, obj.coursesOfStudy, ss, obj.progressOverrides);` 

Is there a reason adding a road is not done by the $store?

src/stores/courseData.js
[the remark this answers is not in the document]
georgiashay: Yeah let's leave that for another PR

georgiashay:
I looked over the cookies/banner stuff and tested it and didn't find any problems after adding the ignoreSet option.

src/components/Auth.vue
```diff
@@ -393,8 +421,8 @@ export default {
       return newRoadData;
     },
     updateRemote: function (roadID) {
-      const newRoad = Object.assign(this.roads[roadID], { id: roadID, override: true, agent: this.getAgent() });
-      this.postSecure('/sync/sync_road/', newRoad)
+      this.$store.commit('updateRoad', roadID, { id: roadID, override: true, agent: this.getAgent() });
```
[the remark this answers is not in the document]
georgiashay: The target and source were in wrong order.  The {} should be first and the road second.

src/components/Auth.vue
```diff
@@ -171,7 +183,7 @@ export default {
         } };
         return this.verify()
           .then(function (verifyResponse) {
-            return !!params
+            return params
```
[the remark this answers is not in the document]
ttpcodes: Hmm, I don't remember explicitly removing the two exclamation marks. Must've been by accident.

src/components/ImportExport.vue
ttpcodes: Another thing I missed, will send this up with the other fixes.

ttpcodes: Just kidding, I remember now. `addRoad` needs access to the `Auth` `ref`, which we can't retrieve from the `Store`, hence why I left it as is. It also doesn't make much sense to move `newRoads` out into the `Store` just for this one case.

src/components/Auth.vue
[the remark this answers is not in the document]
ttpcodes: Oh, this was removed by `ESLint`. So there's actually a `ternary` operator here and we're not just returning the `boolean`. The `ternary` operator which is using `params` is already casting it to `boolean` using what (should be) the same logic, hence it redundancy and why ESLint deleted it.

ttpcodes:
Just went through the comments and made the necessary changes. I also rebased this off of `master` so that this is mergable again.

georgiashay:
The only thing left of concern before merging this was speed.  You have to wait ~2-3 times longer for saving to happen, and you can't do anything while it's saving.  I have identified the source of the problem.  The commit for setRoadProp takes 150 ms, which adds up when you have to process ~15 of them in quick succession.  Looking into this but not too familiar with stores/commits so there might be something obvious I'm missing.

georgiashay:
Looks like it is the Vue.set that is making it take so long

georgiashay:
Nevermind it's not the Vue.set it's just setting the property by any means.

georgiashay:
Since this problem doesn't seem to be introduced by vuex, and it's just an effect of fixing a bug, I think we can go ahead and approve.  I don't want to deploy until we have the speed issues worked out though, so I think maybe I should base my speed branch off of this one and merge them together.  Thoughts?

src/components/Auth.vue
```diff
@@ -215,7 +227,6 @@ export default {
                 if (roadData[r].data.file.contents.progressOverrides === undefined) {
                   roadData[r].data.file.contents.progressOverrides = {};
                 }
-                this.$emit('set-road', roadIDs[r], roadData[r].data.file);
```
npfoss: not sure, but this might be related to the road rename bug?

src/pages/MainPage.vue
```diff
@@ -304,46 +243,35 @@ export default {
...
-          agent: '',
-          contents: {
-            coursesOfStudy: ['girs'],
-            selectedSubjects: [],
-            progressOverrides: {}
-          }
-        }
-      },
+      // TODO: Really we should grab this from a global datastore
+      // now in the same format as FireRoad
```
npfoss: these comments can probably be removed?

src/stores/courseData.js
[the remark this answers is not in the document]
npfoss: created #265 so we don't forget

src/stores/courseData.js
```diff
@@ -0,0 +1,271 @@
...
+  strict: process.env.NODE_ENV !== 'production',
+  state: {
+    activeRoad: '$defaultroad$',
+    addingFromCard: false,
+    classInfoStack: [],
+    cookiesAllowed: undefined,
+    genericCourses: [],
+    genericIndex: {},
+    itemAdding: undefined,
+    // note for later: will need to use Vue.set on roads for reactivity once they come from fireroad
```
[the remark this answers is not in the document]
georgiashay: Yeah I believe we are using Vue.set where applicable now

PR #231, opened 2019-04-30
Banner ui

georgiashay:
the hacky things with emitting cookie events between MainPage and Auth will go away when Vuex is merged

adds colored banners and automatically allows cookies if you edit your road


src/pages/MainPage.vue
```diff
@@ -373,7 +375,9 @@ export default {
     roads: {
       handler: function (newRoads, oldRoads) {
         this.justLoaded = false;
-        if (this.activeRoad !== '') {
+        this.cookiesAllowed = true;
+        this.$refs.authcomponent.allowCookies();
+        if (this.activeRoad != '') {
```
npfoss: is this supposed to be a single `=`?

georgiashay: No, I think this happened because this branch started before the linter.  Will fix when I have time to get to my computer.

georgiashay:
Deployed to dev and noticed a bug caused by the main page vs auth cookies.  Not deploying to prod until fixed (if you opt out then change road, it will update cookies to true - should not do this)

npfoss:
Ah good catch, thanks for being so thorough

georgiashay:
Fixed in #235

Issue #208, opened 2019-04-24
add to about page:


npfoss:
- [x] Thank the Office of the First Year (Kate)
- [x] say something about it always being a work in progress, so more features are coming and feel free to help make them

wishdasher:
Should I also add something about SIPB?

wishdasher:
Done with #233

npfoss:
Sorry for being unresponsive. I agree with all of your choices, thanks for doing this

PR #233, opened 2019-04-30
Update about page

wishdasher:
I thought about adding links to OOFY and SIPB but decided it was too many links everywhere. People should be able to find those easily if they are really curious.
#208


georgiashay:
Looks good to me

PR #235, opened 2019-04-30
Fix banner cookies disallow

georgiashay:
… and only set cookies to true if cookies hasn't been set yet

authCookiesAllowed wasn't being set to false.  This is redundant code but it'll go away with vuex which is happening soon so I'm not going to worry too much about that.

### DIFF
--- a/src/components/Auth.vue
+++ b/src/components/Auth.vue
@@ -447,6 +447,9 @@ export default {
       }
       this.setTabID();
     },
+    disallowCookies() {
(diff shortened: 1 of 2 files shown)


georgiashay:
test looked good to me except I noticed that the loading icon is gone while subjects are loading now.  In class now but can look at this later unless this was intentional and I missed it

georgiashay:
seems like whatever it is happened before these commits, so I'll go ahead and deploy

PR #237, opened 2019-05-02
Subjects with prereq

georgiashay:
Still working on my P1 but was making my road and frustrated this wasn't a feature yet.

- Adds another subject scroller in the class info card with subjects that have given subject as a prereq

Does not:
- Check for equivalence pairs/sets
(Unclear how these should be handled - should we show classes with 6.00 as a prereq under 6.0001? )
- Check if prereq is necessary or one of many you can take (I think we should show it in either case anyway)

### DIFF
--- a/src/components/ClassInfo.vue
+++ b/src/components/ClassInfo.vue
@@ -232,8 +232,9 @@ export default {
     subjectsWithPrereq: function() {
       const currentID = this.currentSubject.subject_id;
       const currentDept = currentID.substring(0,currentID.indexOf("."));
+      var IDMatcher = new RegExp('(?<!\\d)'+currentID.replace(/[.*+?^${}()|[\]\\]/g, '\\$&') + '(?![\\da-zA-Z])');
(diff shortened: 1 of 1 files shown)


src/components/ClassInfo.vue
```diff
@@ -224,6 +228,23 @@ export default {
...
         : { reqs: [] };
+    },
+    subjectsWithPrereq: function() {
+      const currentID = this.currentSubject.subject_id;
+      const currentDept = currentID.substring(0,currentID.indexOf("."));
+      var IDMatcher = new RegExp('(?<!\\d)'+currentID.replace(/[.*+?^${}()|[\]\\]/g, '\\$&') + '(?![\\da-zA-Z])');
+      return this.subjects.filter(function(s) {
+        return s.prerequisites !== undefined && IDMatcher.test(s.prerequisites);
+      }).sort(function(s1, s2) {
+        //show subjects in same department first
```
npfoss: This is a nice touch!

PR #244, opened 2019-05-04
expand list of known issues

npfoss:


georgiashay:
UI quirks seems kinda vague, since we only have one thing maybe we want to just call it search box problems or something like that?  Other than that looks good.

PR #250, opened 2019-05-06
Unscope style and make css selector more specific so it wins

### LINKED ISSUE #245: Plus button gets pushed off screen again
Some recent commit must have caused this since this problem has been fixed once before.

image

georgiashay:
Fixes #245 

A little wary about unscoping the style but there is no other easy way to select the generated html.

### DIFF
--- a/src/components/RoadTabs.vue
+++ b/src/components/RoadTabs.vue
@@ -151,9 +151,9 @@ export default {
 };
 </script>
 
-<style scoped>
+<style>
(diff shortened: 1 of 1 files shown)


georgiashay:
I actually ended up including this in the vuex branch since it was basically impossible to test without.

I think unscoping is fine as long as we scope anything in the same component that isn't generated html selectors.

npfoss:
ah ok. I'll just add a warning in the comments then

PR #258, opened 2019-05-18
Search placement card

### LINKED ISSUE #216: wacky search placement
On the dev site (the errors are unrelated, mentioned in #215):
image

On localhost (note the offset):
image

They both went away upon resizing the screen.

The offset happens when you reload with inspect element open, but I have no idea how to reproduce the first one..

georgiashay:
Fixes #216 

Got rid of v-menu and used v-card instead. 

Replaced updating position on size with @media.  It seems like there are only 2 sizes it is possible to get to, but maybe we should test this on a bigger screen to make sure?  If we think this is too risky we can add the repositioning back in, but I am hesitant to keep that lag.

### DIFF
--- a/src/pages/MainPage.vue
+++ b/src/pages/MainPage.vue
@@ -86,27 +86,6 @@
         />
       </v-layout>
 
-      <v-menu
-        :close-on-content-click="false"
-        v-model = "searchOpen"
-        :position-x = "searchX"
-        :position-y = "searchY"
-      >
-        <class-search
-          id="searchMenu"
-          ref="searchMenu"
-          class="search-menu"
-          :search-input="searchInput"
-          :subjects="subjectsInfo"
-          :generic-courses="genericCourses"
-          :class-info-stack="classInfoStack"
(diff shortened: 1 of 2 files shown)


src/pages/MainPage.vue
```diff
@@ -420,18 +411,6 @@ export default {
...
-    this.searchY = $("#searchInputTF").offset().top + $("#searchInputTF").outerHeight();
-
-    $(window).on("resize", function() {
-      this.searchX = $("#searchInputTF").offset().left;
-      this.searchY = $("#searchInputTF").offset().top + $("#searchInputTF").outerHeight();
-    }.bind(this));
-
-    document.body.addEventListener('click', function (e) {
-      this.showSearch = false;
-    }.bind(this));
```
npfoss: I think we want this -- closes search when click away

georgiashay:
Fixed problems, also added close on esc (although it only works if the input is focused, so won't work if you have dragged something from the class search and your focus isn't the search anymore).

PR #260, opened 2019-05-18
Make drag message visible again

### LINKED ISSUE #259: Drag class message not being shown
The messages for when you drag classes into each semester that explain the color are gone

georgiashay:
Fixes #259


georgiashay:
Possible separate issue: on a smaller screen, the message can be two lines long, making it jump when you drag something across it.

PR #263, opened 2019-05-30
Partial subjects

### LINKED ISSUE #262: Improved handling of partial subjects
FYI that I've implemented a better way to handle partial subjects (for example, 6.0001 + 6.0002 = 6.00). This was completed for the frontend's prereq computation in #135, I believe - but back then we had to hardcode those subjects, which wasn't ideal. Now each subject can have optional `parent` and `children` keys to specify which courses can add together to satisfy another course. For example, 6.0001 and 6.0002 each have `"parent": "6.00"`, and 6.00 has `"children": ["6.0001", "6.0002"]`. Other examples include 6.041A/B and 3.016A/B.

You can see how I've implemented the requirement checking [here](https://github.com/venkatesh-sivaraman/fireroad-server/blob/885a9a67fa48113e733b7f9f5cd000684bc6d69c/catalog/models.py#L428). This feature is currently on dev, and will soon move to prod.

### LINKED ISSUE #270: Equivalent subjects don't count for prereqs
If a subject is equivalent to a subject needed for prereq, it should count for that prereq

Affected subjects include ESG subjects not fulfilling prereqs

georgiashay:
Fixes #262 
Fixes #270 

Uses new parent and children keys on subjects instead of hardcoded equivalence pairs (6.0001 + 6.0002 = 6.00)

Also no longer includes things like 'permission of instructor' as automatic true, so you have to explicitly hide warnings if this is how you are able to take the course.

### DIFF
--- a/src/components/Semester.vue
+++ b/src/components/Semester.vue
@@ -288,6 +288,11 @@ export default {
         // subj not found in known courses
         return false;
       }
+
+      if(subj.equivalent_subjects !== undefined && subj.equivalent_subjects.indexOf(req) >= 0) {
+        return true;
+      }
+
       //ex: 6.00 satisfies the 6.0001 requirement
       if (subj.children !== undefined && subj.children.indexOf(req) >= 0) {
         return true;
--- a/src/components/Semester.vue
+++ b/src/components/Semester.vue
@@ -319,7 +319,15 @@ export default {
       const _this = this
       for (let i = 0; i < splitReq.length; i++) {
         if (splitReq[i].indexOf('"') >= 0) {
-          splitReq[i] = 'true';
+          //don't include things like 'permission of instructor'
+          splitReq[i] = '';
+          //remove operator before & parentheses if directly before and after
+          if(splitReq[i] !== '(') {
+            splitReq[i-1] = '';
+          } else if(splitReq[i+1] === ')') {
+            splitReq[i-1] = '';
+            splitReq[i+1] = '';
+          }
         } else if ('()/, '.indexOf(splitReq[i]) < 0) {
           if (allIDs.indexOf(splitReq[i]) >= 0) {
             splitReq[i] = 'true';
--- a/src/components/Semester.vue
+++ b/src/components/Semester.vue
@@ -331,7 +331,7 @@ export default {
       if (subj.parent !== undefined && req === subj.parent) {
         let parentCourse = this.allSubjects[this.$store.state.subjectsIndex[subj.parent]];
         if (parentCourse !== undefined) {
-          if (parentCourse.children.reduce((acc, sid) => acc && allSubjects.indexOf(sid) >= 0, true)) {
+          if (parentCourse.children.map(sid => allSubjects.indexOf(sid) >= 0).every()) {
             return true;
           }
         }
@@ -369,7 +369,7 @@ export default {
           if (allIDs.indexOf(splitReq[i]) >= 0) {
             splitReq[i] = 'true';
           } else {
-            const anyClassSatisfies = subjects.map((s) => _this.classSatisfies(splitReq[i], s.id, allIDs)).reduce((a, b) => a || b, false);
+            const anyClassSatisfies = subjects.map((s) => _this.classSatisfies(splitReq[i], s.id, allIDs)).some();
             splitReq[i] = anyClassSatisfies ? 'true' : 'false';
           }
         }
--- a/src/components/Semester.vue
+++ b/src/components/Semester.vue
@@ -318,18 +318,18 @@ export default {
         return false;
       }
 
-      if(subj.equivalent_subjects !== undefined && subj.equivalent_subjects.indexOf(req) >= 0) {
+      if (subj.equivalent_subjects !== undefined && subj.equivalent_subjects.indexOf(req) >= 0) {
         return true;
       }
 
-      //ex: 6.00 satisfies the 6.0001 requirement
+      // ex: 6.00 satisfies the 6.0001 requirement
       if (subj.children !== undefined && subj.children.indexOf(req) >= 0) {
         return true;
       }
 
-      //ex: 6.0001 and 6.0002 together satisfy the 6.00 requirement
+      // ex: 6.0001 and 6.0002 together satisfy the 6.00 requirement
       if (subj.parent !== undefined && req === subj.parent) {
-        let parentCourse = this.$store.state.subjectsInfo[this.$store.state.subjectsIndex[subj.parent]];
+        const parentCourse = this.$store.state.subjectsInfo[this.$store.state.subjectsIndex[subj.parent]];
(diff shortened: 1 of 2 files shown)


src/components/Semester.vue
```diff
@@ -319,21 +322,21 @@ export default {
...
       for (let i = 0; i < splitReq.length; i++) {
         if (splitReq[i].indexOf('"') >= 0) {
-          splitReq[i] = 'true';
-        }
-        if ('()/, '.indexOf(splitReq[i]) < 0) {
+          //don't include things like 'permission of instructor'
+          splitReq[i] = '';
+          //remove operator before & parentheses if directly before and after
+          if(splitReq[i] !== '(') {
+            splitReq[i-1] = '';
```
npfoss: This feels a little dangerous, and I haven't come across an example (in a short search) where it's not already an empty string. Why do we need this particular case?

src/components/Semester.vue
```diff
@@ -319,21 +322,21 @@ export default {
...
+          } else if(splitReq[i+1] === ')') {
+            splitReq[i-1] = '';
+            splitReq[i+1] = '';
```
npfoss: This one also feels dangerous, like it seems like this could potentially lose information about the difference between (A and B) and (A/B) or something similar? Why were these added? I'm sure it's a good reason I just don't understand the format this is coming in as well.

src/components/Semester.vue
```diff
@@ -319,21 +322,21 @@ export default {
...
-            const anyClassSatisfiesAlone = subjects.map((s) => _this.classSatisfies(splitReq[i], s.id)).reduce((a, b) => a || b, false);
-            let anyClassesSatisfyTogether = false;
-            for (let e = 0; e < EQUIVALENCE_SETS.length; e++) {
-              if (EQUIVALENCE_SETS[e][1] === splitReq[i] && EQUIVALENCE_SETS[e][0].reduce((acc, sid) => acc && allIDs.indexOf(sid) >= 0, true)) {
-                anyClassesSatisfyTogether = true;
-                break;
-              }
-            }
-            splitReq[i] = (anyClassSatisfiesAlone || anyClassesSatisfyTogether) ? 'true' : 'false';
+            const anyClassSatisfies = subjects.map((s) => _this.classSatisfies(splitReq[i], s.id, allIDs)).reduce((a, b) => a || b, false);
```
npfoss: clearer with `some` than `reduce` I think

src/stores/courseData.js
```diff
@@ -171,7 +171,10 @@ const store = new Vuex.Store({
       }
     },
     removeClass (state, classInfo) {
+      console.log(classInfo);
+      console.log(state.roads[state.activeRoad].contents.selectedSubjects);
       const classIndex = state.roads[state.activeRoad].contents.selectedSubjects.indexOf(classInfo);
+      console.log(classIndex);
```
npfoss: FYI you probably want to remove this before merging master in, it caused some conflicts when I did that locally. (and ofc before merging the PR too)

npfoss:
Does this fix #270?

src/components/Semester.vue
georgiashay: Unfortunately it's been a while and I don't 100% understand my comments either but I think what this is doing is 
1)  if the part right before it isn't an open parenthesis it must be an operator (',' or '/'), so that should be removed (set to '') since the string is being removed
2) if there are parenthesis before and after, remove both of them

An example I made would be "1.01/2.003/(4.05,6.06,7.07,\"permission of instructor\")"
In this case there is a comma before "permission of instructor" and a parenthesis after.  The comma and string should be removed so it becomes "1.01/2.003/(4.05,6.06,7.07)"

src/components/Semester.vue
georgiashay: General reasoning for this part of the code is in the other comment

Not sure what you mean by the difference between (A and B) and (A/B)?  (A and B) should be written as (A,B) but I don't know what you mean by losing information

georgiashay:
It actually looks like #271 should have fixed it by itself?  I think both branches fix it independently

src/components/Semester.vue
npfoss: ah I see now, I didn't get what was happening before. thanks

src/components/Semester.vue
npfoss: yep I was being dumb sorry

src/components/Semester.vue
```diff
@@ -348,21 +356,21 @@ export default {
       const _this = this;
       for (let i = 0; i < splitReq.length; i++) {
         if (splitReq[i].indexOf('"') >= 0) {
-          splitReq[i] = 'true';
-        }
-        if ('()/, '.indexOf(splitReq[i]) < 0) {
+          //don't include things like 'permission of instructor'
+          splitReq[i] = '';
+          //remove operator before & parentheses if directly before and after
+          if(splitReq[i] !== '(') {
```
npfoss: is this line a typo? we just set `splitReq[i]` to the empty string, so it will always remove the thing before. also the `else if` will never happen?

georgiashay: Definitely a typo but (I think it should be i-1) but now I'm looking back on that whole section again and I'm not sure it actually works?  Regardless I didn't leave clear enough comments to come back to it and understand what's going on so I think I'm going to have to redo this part

npfoss: I think this makes more sense. But would also fail the `("permission of instructor")` case now?

Either way, might not be worth bothering... see comment below :P

georgiashay:
That's not exactly what it's doing.  If it's working correctly it should be cutting the `"permission of instructor"`s out entirely, which is effectively making it false for an or statement or true in an and statement.  I originally had them either default true or default false but either is bad for the user since default true will obscure missing prereqs that you have, and default false will make it impossible to meet the prereqs.

src/components/Semester.vue
georgiashay: I looked over the prereqs/coreqs returned from fireroad and I didn't see any of those cases with parenthesis around just `"permission of instructor"`, but now that I think about it there were several cases with parenthesis around strings that would all get cut out (e.g. `("placement exam", "permission of instructor")` would become `()` which isn't evaluable.  It would be easy enough to just cut out all cases of `()` though.

npfoss:
Isn't that what we want, though? Like if the only way you have the prereqs is if permission of instructor is true, then it should totally show a warning you have to manually check and dismiss, right?

georgiashay:
I think that does make more sense.

georgiashay:
This could get confusing on classes with lots of prereqs if you think you are overriding just the 'permission of instructor part' but there is another missing requirement since we don't tell the user which prereqs are missing.  Might be a good feature to add later in the class info card at least to put check marks or x's next to requirements?

npfoss:
ooh yeah that's a good idea

PR #266, opened 2019-06-16
add undefined checks

npfoss:
I was getting a bunch of errors from at least one of these lines, added a safeguard everywhere.

Can't reproduce it any more on master though so I'm leaving this PR unmerged for someone else to confirm it seems logical, since I think this must be a rare error.

### DIFF
--- a/src/components/Semester.vue
+++ b/src/components/Semester.vue
@@ -103,7 +103,7 @@ export default {
         const subjectWarnings = [];
         const subjID = this.semesterSubjects[i].id;
         let subj;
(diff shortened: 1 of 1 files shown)


georgiashay:
Hmm, I'm not sure why this should be necessary since the subjectsIndex starts out as [] not undefined.  Is it possible for Semester.vue to load before the store?  I also can't reproduce on master right now.

npfoss:
Oh hm you're probably right then, what I saw must have been caused by something else...

I can't reproduce it either so it might have been a strange transient hot reload issue or something?

PR #269, opened 2019-06-24
Upgraded axios

georgiashay:


package-lock.json
```diff
@@ -3987,6 +4016,7 @@
           "version": "3.2.6",
           "resolved": "https://registry.npmjs.org/debug/-/debug-3.2.6.tgz",
           "integrity": "sha512-mel+jf7nrtEl5Pn1Qx46zARXKDpBbvzezse7p7LqINmdoIk8PYP5SySaxEmYv6TZ0JyEKA1hsCId6DIhgITtWQ==",
+          "dev": true,
```
npfoss: Are all of these `"dev": true`s supposed to be here? Does it even matter since it all gets compiled anyways?

PR #271, opened 2019-07-23
Prereq equivalent subjects

georgiashay:
Class satisfies requirement if the requirement is the name of any equivalent subject

### DIFF
--- a/src/components/Semester.vue
+++ b/src/components/Semester.vue
@@ -288,6 +288,11 @@ export default {
         // subj not found in known courses
         return false;
       }
+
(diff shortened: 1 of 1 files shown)


src/components/Semester.vue
```diff
@@ -288,6 +288,11 @@ export default {
         // subj not found in known courses
         return false;
       }
+
+      if(subj.equivalent_subjects !== undefined && subj.equivalent_subjects.indexOf(req) >= 0) {
```
npfoss: any reason not to use `includes`?

PR #272, opened 2019-07-23
Updated lodash

georgiashay:
Fixed high severity security alert

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -101,6 +101,12 @@
             "lodash": "^4.17.11",
             "to-fast-properties": "^2.0.0"
           }
+        },
+        "lodash": {
+          "version": "4.17.15",
+          "resolved": "https://registry.npmjs.org/lodash/-/lodash-4.17.15.tgz",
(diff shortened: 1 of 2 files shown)


npfoss:
seems like this has broken something so I fixed it on the axios-security PR instead (#269)
image

npfoss:
actually I'm dumb and the other one has the same problem. so let's figure that out on this PR

georgiashay:
That looks like the regex error that's always been there

georgiashay:
Although on second thought 'fa' seems like a valid regex?

georgiashay:
This might not be related to lodash but we should create another issue about invalid regexes only searching by plain text

npfoss:
idk it wasn't happening on master when I sat down 30 mins ago... and it's not on prod either. can't investigate now though gtg sorry. I'll be back tonight

georgiashay:
What steps can I take to recreate the issue?

npfoss:
right now for me if I `npm install` (this branch or master) and then click on anything in search results I get those errors

I'm working on it now though, just give me a bit

npfoss:
update: it's a problem with lookbehinds, not this. merging

PR #275, opened 2019-07-25
fix regex bug

npfoss:
Lookbehind is allowed in Chrome but not Firefox


npfoss:
@georgiashay can you take a look at this one please? I edited the offending line to unbreak it and left the original commented out, but you know this logic best and I'm not sure what exactly it's doing

PR #282, opened 2019-08-09
remove all lint errors

npfoss:
Things I wasn't sure about that need taking care of:
- [x] `Array prototype is read only, properties should not be added  no-extend-native` in `ConflictDialog.vue`: should we disable this or is that actually bad practice?
- [x] `Unexpected side effect in "autocomplete" computed property  vue/no-side-effects-in-computed-properties` in `ClassSearch.vue`: this does seem like bad practice.

Also almost every single warning is about defining prop types, so that would probably be straightforward to tackle in a subsequent PR if someone wants to take it on. I didn't do it here because the errors took long enough.

### DIFF
--- a/src/app.js
+++ b/src/app.js
@@ -1,3 +1,4 @@
+/* eslint-disable no-new */
 import Vue from 'vue';
 import App from './App.vue';
 import MainPage from './pages/MainPage.vue';
--- a/src/components/Auth.vue
+++ b/src/components/Auth.vue
@@ -46,7 +46,6 @@
 <script>
 import axios from 'axios';
 import moment from 'moment';
-import Vue from 'vue';
 import UAParser from 'ua-parser-js';
 import simpleSSMixin from './../mixins/simpleSelectedSubjects.js';
 
@@ -179,7 +178,7 @@ export default {
             return verifyResponse.data;
           } else {
             this.logoutUser();
-            return Promise.reject('Token not valid');
+            return Promise.reject(new Error('Token not valid'));
           }
         }.bind(this));
     },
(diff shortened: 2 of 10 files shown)


georgiashay:
Extending array prototype probably is bad practice and we only need those functions for certain arrays anyway.  Our use case doesn't fit in the [reasons why you would disable the rule](https://eslint.org/docs/rules/no-extend-native).

For the autocomplete side effect, I think I remember having trouble with a related problem earlier.  I tried tying them together like this:
`'chosenFilters.nameInput': function() {
      return this.searchInput;
    }`
in a computed function but it did not ever call it.  Not sure how to fix the issue

npfoss:
I spent so long debugging that but eventually figured out it wasn't being reactive because the v-if around the search card disappears/appears when the input goes between empty and nonempty, and when a component comes into existence or disappears, the reactive things attached to its props don't get triggered. Changing the v-if to a v-show make the component exist all the time, but be hidden when the condition is false, thus solving the reactivity thing. Jeez.

npfoss:
Please take a close look at those last two commits @georgiashay and make sure my changes shouldn't change any behavior. These ones were more substantial than simple style adjustments.

georgiashay:
I'm seeing that now the search bar juts out a little on page load.  It fixes itself when resized.  Not seeing this behavior on master and I don't think I saw it before the rebase but I can't be sure since I wasn't looking for it.  
image

npfoss:
Oops, good catch, thanks. The problem was that with v-show the component was being created and mounted on pageload (instead of every time it opens), before the inspector's window size modification was taken into account, I guess. Should be fixed now.

PR #287, opened 2019-08-20
show warnings and stuff if subject not offered this year

### LINKED ISSUE #286: not-offered-year is broken
24.118 isn't offered this year, and it says so in `subjectsInfo`, but we aren't displaying a warning:
image

npfoss:
fixes #286

### DIFF
--- a/src/components/ClassInfo.vue
+++ b/src/components/ClassInfo.vue
@@ -58,12 +58,16 @@
               </v-btn>
             </v-layout>
             <h4 v-if="currentSubject.is_historical">
-              <v-icon>warning</v-icon>
+              <v-icon small>
+                warning
+              </v-icon>
               This subject is no longer offered (last offered {{ currentSubject.source_semester.split("-").join(" ") }})
             </h4>
-            <h4 v-else-if="lateSchedule(currentSubject, $store.state.genericIndex)">
-              <v-icon>warning</v-icon>
-              This subject does not have a schedule yet, so it may not be offered next semester.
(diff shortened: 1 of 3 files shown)


georgiashay:
Maybe we should also update the class info card because the "does not have a schedule yet" warning isn't really necessary if it's not offered next year.

npfoss:
Oh, good call. I'll add that when I get the chance

npfoss:
Ok I added that, and made the icons small because I think it looks better.

I also removed the no schedule warning from the class info card because that feels like a very situational warning, (i.e. only matters when it's in your schedule for next semester), whereas this one is the kind you like to know about globally, e.g. the classInfoCard context. Thoughts?

georgiashay:
I think the smaller icon looks good (although there might be too much space above the warning?)

I agree about removing the no schedule warning from the class info card.  

I think there's an issue with current semester/scheduled semester that is affecting the warnings, though.  Since it's not September yet it thinks it's the spring semester so it's applying the no schedule warning to spring classes instead of fall classes.  Perhaps worse, it thinks it's Spring 2020 not Spring 2019 (at least according to what is displayed) so it's also wrong at being wrong.  Might be a separate issue or worth addressing here if we want to consider the semester start/end cutoffs differently for different purposes.

npfoss:
I couldn't reproduce this for a while, and it's because it was only happening in the first two semesters before you manually set class year. The culprit turned out to be this line:
```js
this.currentSemester = (month >= 4 && month <= 10) ? 3 : 1;
```
Which is backwards. I changed it to `1 : 3`.

PR #302, opened 2019-09-12
Unit testing

georgiashay:
Adds jest unit testing
Adds example unit test file for Conflict Dialog
Fixes some bugs (this is why we need unit testing, even on the simplest component) and adds some helpful id's to conflict dialog
Adds unit testing to travis (might work?)


georgiashay:
i think i broke travis :( will figure out why

georgiashay:
revived travis successfully

georgiashay:
To clarify:
* this added scaffolding for unit testing but only actual unit tests for conflict dialog as a proof of concept.  scaffolding can be discussed - we might want a test folder for each component or something like that rather than one big test folder, but this should be easy to change later
* Travis CI now has two stages, lint and unit
* In addition to retroactively adding code coverage, we should add unit tests for things we add from now on - in particular when we add petitions

georgiashay:
also i apologize this is the ugliest diff i've ever seen.  ignore everything under coverage, the real changes happened in tests/unit/,  ConflictDialog.vue, package.json, jest.config.js, travis.yml and maybe a few other places

.babelrc
```diff
@@ -1,10 +1,5 @@
 {
-  "presets": [
-    ["@babel/preset-env", {
-      "modules": false,
-      "targets": {
-        "browsers": ["> 1%", "last 2 versions", "not ie <= 8"]
-      }
-    }]
```
npfoss: why did this change? (I don't actually know what it's for)

package.json
```diff
@@ -2,29 +2,17 @@
...
-  "keywords": [
-    "mit",
-    "courseroad"
-  ],
-  "author": "npfoss, cvorbach, ksmori",
-  "license": "MIT",
-  "bugs": {
-    "url": "https://github.com/npfoss/courseroad2/issues"
+    "dev": "webpack-dev-server --env.APP_URL=http://localhost:8080 --env.FIREROAD_URL=https://fireroad-dev.mit.edu --mode=development --config build/webpack.config.dev.js",
+    "lint-fix": "node_modules/.bin/eslint --ext .js,.vue --fix src/",
```
npfoss: why did you move these around? I feel like the two lint commands should be together

package.json
```diff
@@ -2,29 +2,17 @@
...
+    "test:unit": "vue-cli-service test:unit --coverage"
```
npfoss: is there a reason not to replace the existing `test` command (which doesn't do anything) with this one?

tests/unit/.eslintrc.js
```diff
@@ -0,0 +1,5 @@
+module.exports = {
+  env: {
+    jest: true
+  }
+}
```
npfoss: should this not be the same .eslintrc as in the outer project? why have two?

.babelrc
ttpcodes: From my understanding, `preset-env` is used to convert some of our JavaScript code written for `courseroad2` so that it actually works in browsers (some later JavaScript features vary greatly in terms of browser support). `transform-es2015-modules-commonjs` is used to convert JavaScript code into `CommonJS` (the module system used for requiring packages in things like `Node.js`.

From my understanding, we need to keep `preset-env` around, and I'm not sure why `transform-es2015-modules-commonjs` is needed unless something in `vue-cli` requires it. @georgiashay Do you know which version of `vue-cli` you're using so I can look into it?

tests/unit/.eslintrc.js
ttpcodes: There might be some rules that might need to be different so the linter doesn't fail on the test code (for example, the linter *might* freak out since we're using keywords like `describe` and `beforeEach` that aren't imported like in other code because the test runner defines these global functions.

We could try to copy the existing `ESLint` configuration here and see if the build freaks out, and then modify the rules accordingly until it passes.

.babelrc
```diff
@@ -1,10 +1,5 @@
...
+  "plugins": [
+    "transform-es2015-modules-commonjs"
   ]
-}
```
ttpcodes: There's a random change here that has to do with a newline character being removed from the end of the file. We should probably decide as a team whether we should enforce newline at the end of a file or not so that we don't have this extraneous change in commits every time someone with a different newline configuration pushes a change to a particular file.

.travis.yml
```diff
@@ -1,6 +1,16 @@
...
 node_js:
-  - "10"
+  - "11"
 
-script: npm run lint
\ No newline at end of file
+stages:
+  - compile
+  - test
+  - deploy
```
ttpcodes: It looks like the `stages` configuration is incorrect. The only stages we have defined are `lint` and `unit`, and all of the stages here do not exist.

Additionally, the `stages` key is only really needed to define the order the stages should run for special purposes (such as prepending some random stage). Can we try to remove this block and see if it breaks anything?

package.json
ttpcodes: `test:unit` is the default name generated by `vue-cli` when adding unit tests. It's to allow for developers to differentiate between calling unit tests and end-to-end tests (which `vue-cli` places under `test:e2e`.

I personally think it would be better to either remove the `test` command from `package.json` or rewrite it so it calls both unit tests and end-to-end tests.

(Speaking of which, where's the `test:e2e` command?)

.babelrc
georgiashay: Added back in after much hassle.  Weird babel errors https://github.com/facebook/jest/issues/6913

package.json
georgiashay: I haven't added e2e yet, this is just unit testing

tests/unit/.eslintrc.js
georgiashay: Copied configuration but changed no-undef to a warning

.travis.yml
georgiashay: Removed block successfully

georgiashay:
Fixed some things, ready for review again

tests/unit/.eslintrc.js
```diff
@@ -0,0 +1,38 @@
+module.exports = {
```
npfoss: wait a second, do we need another eslint here? I don't think so? It should just use the main one?

georgiashay: this was definitely auto-generated, so idk.  in theory it could just use the main one but there are slightly different rules as I removed the errors from no-undef

tests/unit/.eslintrc.js
```diff
@@ -0,0 +1,38 @@
...
+    "no-var": 0,
+    "prefer-const": 1,
+    "no-undef": 1
+  },
+  "overrides": [
+    {
+      "files": [
+        "**/__tests__/*.{j,t}s?(x)"
+      ],
+      "env": {
```
ttpcodes: Is there a reason why this part of the configuration is under `overrides` and isn't just outside with everything else?

georgiashay: It's copied from the lintrc file, so im not sure why it is this way but I didn't touch it

npfoss:
Update from the meeting, for the record: Georgia is trying to figure out this linting thing, apparently there might be an eslint plugin or something. Worst case, we'll remove all eslintrc files except the top level one and put something there with a special rule for the tests directory (although this solution should make us a little sad).

tests/unit/.eslintrc.js
ttpcodes: So apparently, under [this documentation](https://eslint.org/docs/user-guide/configuring) under the section `Configuration Cascading and Hierarchy`, it turns out that `ESLint` actually inherits previous `.eslintrc.js` files and attempts to merge the contents of the `.eslintrc.js` file in the current directory with any other encountered ones. 

This actually wasn't a necessary change then. @georgiashay Can this be restored to the original that was there before? Assuming I'm not remembering incorrectly, this should be the original contents of this `.eslintrc.js`:
```
module.exports = {
    env: {
      jest: true
  }
}

georgiashay: This wasn't working for some reason, but I found the culprit:  For some reason, using extends: standard was messing with the jest linting plugin.  Do we need this, or could we just add the specific rules that we want?

georgiashay: nevermind it just doesn't work

georgiashay:
Condensed into one eslintrc file; use jest plugin for eslint and globals from setup.js for unit testing files

PR #305, opened 2019-09-13
Auto set user year2

miriam-rittenberg:
This should load the user's year and keep the year stored, and it should change the base year of the change your year dialog. I also noticed a bug in changeSemester that prevented the dialog from changing the year while you are not logged in. I fixed that but couldn't figure out how to stop the year from resetting to freshman when you reload the page.


src/stores/courseData.js
```diff
@@ -240,6 +241,10 @@ const store = new Vuex.Store({
     setSubjectsInfo (state, data) {
       state.subjectsInfo = data;
     },
+    setUserYear (state, year) {
+      console.log('set user year to', year);
```
npfoss: Please remove logs before merging

miriam-rittenberg: I'm mostly storing year because I need the year in a few places and this requires less converting back and forth. But maybe I should move `currentSemester` to the store and just calculate the year when I need it.

miriam-rittenberg: Also I'm not actually sure how to convert from year to semester, which semester is 1/2/etc?

miriam-rittenberg: Georgia explained this at the meeting, I will change this to just using Semester

npfoss: sounds good!

miriam-rittenberg:
Login started working again, and it looks like everything works right

src/components/Road.vue
```diff
@@ -72,18 +71,14 @@ export default {
       const today = new Date();
       const currentYear = today.getFullYear();
       const baseYear = (today.getMonth() >= 5 && today.getMonth() <= 10) ? currentYear + 1 : currentYear;
-      return baseYear - Math.floor((this.currentSemester - 1) / 3);
+      return baseYear - Math.floor((this.$store.state.currentSemester - 1) / 3);
```
npfoss: maybe `year` would be best as a computed property in the store now, since this `-1)/3` logic is in multiple places?

src/pages/MainPage.vue
```diff
@@ -332,7 +329,7 @@ export default {
   mounted () {
     const today = new Date();
     const month = today.getMonth();
-    this.currentSemester = (month >= 4 && month <= 10) ? 1 : 3;
+    this.$store.state.currentSemester = (month >= 4 && month <= 10) ? 1 : 3;
```
npfoss: it's better to use a mutation here (`setCurrentSemester`), Vuex doesn't like it when state is directly modified externally

src/stores/courseData.js
```diff
@@ -10,7 +10,7 @@ const DATE_FORMAT = 'YYYY-MM-DDTHH:mm:ss.SSS000Z';
 const store = new Vuex.Store({
   strict: process.env.NODE_ENV !== 'production',
   state: {
-    userYear: 0,
```
npfoss: like I said in an above comment, you could keep all of the stuff which uses `userYear` without duplicating data (the original reason for these changes) by making it a computed property in the store (here). (I think these are called `getters`)

Only if you think that's a good idea though, ofc. you've thought about this more than me at this point

src/components/Road.vue
```diff
@@ -15,10 +15,9 @@
       :semester-subjects="selectedSubjects[index-1]"
       :road-i-d="roadID"
       :is-open="visibleList[index-1]"
-      :base-year="baseYear"
       :adding-from-card="addingFromCard"
       :dragging-over="dragSemesterNum===index-1"
-      @change-year="changeYearDialog = true"
+      @open-change-year-dialog="changeYearDialog = true"
```
npfoss: Thank you for changing this name.

PR #319, opened 2019-10-04
improve courses box UI

npfoss:
removed the icon and need to click, and improved the surrounding text


npfoss:
Yeah I considered that a feature, sort of a slight indicator something is wrong

PR #322, opened 2019-10-07
Better filter code

georgiashay:
Abstracts out filters into classes that can be extended to things other than regex
Implements sorting in a way that can be more easily extended in the future


src/components/ClassSearch.vue
```diff
@@ -52,6 +52,193 @@ import FilterSet from './FilterSet.vue';
...
+  constructor (name, filters, combination) {
+    this.name = name;
+    this.filters = filters;
+    this.combine = {
+      'AND': (a, b) => a && b,
+      'OR': (a, b) => a || b
+    }[combination];
+  }
+
+  matches (subject, applied, inputs) {
```
npfoss: `applied` could have a better name, it's not clear that this is which filters are active (right?)

src/components/ClassSearch.vue
```diff
@@ -133,84 +282,33 @@ export default {
       return this.$store.state.genericCourses.concat(this.$store.state.subjectsInfo);
     },
     autocomplete: function () {
-      // only display subjects if you are filtering by something
-      let returnAny = false;
+      // Only display subjects if you are filtering by something
+      let returnAny = this.nameInput.length > 0;
       for (const filterName in this.chosenFilters) {
-        returnAny = returnAny || this.chosenFilters[filterName].length;
+        returnAny = returnAny || this.chosenFilters[filterName].reduce((acc, applied) => acc || applied, false);
```
npfoss: this reduce can be an `any`

src/components/ClassSearch.vue
```diff
@@ -133,84 +282,33 @@ export default {
...
-          return matches.lastIndexOf(true);
-        };
-        const getOrder = function (subject) {
-          const idMatch = getOrderForString(subject.subject_id);
-          const nameMatch = getOrderForString(subject.title);
-          return Math.max(idMatch, nameMatch);
-        };
+        matches = matches & textFilter.matches(subject, { nameInput: this.nameInput });
+        return matches;
+      }.bind(this));
```
npfoss: This looks like a natural `all(Object.keys(this.allFilters).map(...))` kind of thing, if you feel like it. This way is fine too though if you think it's better.

src/components/ClassSearch.vue
```diff
@@ -52,6 +52,193 @@ import FilterSet from './FilterSet.vue';
 import $ from 'jquery';
 import Vue from 'vue';
 
+class Filter {
+  constructor (name, shortName, filter, attributeNames, requires, mode) {
+    this.name = name;
+    this.short = shortName;
+    this.filter = filter;
+    this.attributes = attributeNames;
+    this.requires = requires;
```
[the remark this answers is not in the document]
georgiashay: I should actually remove it from the main class since it's only used in the regex filter; but will do

src/components/ClassSearch.vue
npfoss: much better, thanks!

georgiashay:
This seems to be less responsive than on master.  Not sure we should merge if it's noticeably slower

npfoss:
> less responsive than master

Are you sure? It still seems pretty snappy to me...

src/components/ClassSearch.vue
```diff
@@ -52,6 +50,204 @@ import FilterSet from './FilterSet.vue';
...
+    var testFunction = RegexFilter.getRegexTestFunction(regex);
+    super(name, shortName, testFunction, attributeNames, mode);
+    this.regex = regex;
+    // Input required to construct the regex (for filters that change based on input)
+    this.requires = requires;
+  }
+
+  static escapeRegex (regex) {
+    return regex.replace(/[.*+?^${}()|[\]\\]/g, '\\$&');
+  }
```
npfoss: I appreciate this function

georgiashay:
It seemed like a noticeable difference to me but I can test again

georgiashay:
There's still a difference in speeds but I think it's no longer perceptible? (I'm seeing like 55ms on master, 85ms here for ..a)

src/components/ClassSearch.vue
```diff
@@ -52,6 +50,207 @@ import FilterSet from './FilterSet.vue';
...
+    } catch (e) {
+      // If a regex cannot be constructed, default to matching the literal
+      regexObject = new RegExp(RegexFilter.escapeRegex(regex), 'i');
+    }
+    // Regex test function only works when bound to the regex object
+    return regexObject.test.bind(regexObject);
+  }
+
+  setupInputs (inputs) {
+    if (this.requires != undefined) {
```
npfoss: should this be `!==`?

src/components/ClassSearch.vue
```diff
@@ -52,6 +50,207 @@ import FilterSet from './FilterSet.vue';
...
+  matches (subject, active, inputs) {
+    if (!active.some((a)=>a)) {
+      return true;
+    }
+    const baseCombinationValue = !this.combine(true, false);
+    var isMatch = baseCombinationValue;
+    for (var f = 0; f < this.filters.length; f++) {
+      if (active[f]) {
+        isMatch = this.combine(isMatch, this.filters[f].matches(subject));
+        if(isMatch != baseCombinationValue) {
```
npfoss: should this be `!==`?

georgiashay:
Yep, my bad I missed a few !== -  fixed now

PR #323, opened 2019-10-09
make buttons not rounded

npfoss:
Do you agree?


georgiashay:
I think I do like them better this way but they need less padding now

npfoss:
I made them smaller but I'm not sure if that's better or worse. What do you think?

npfoss:
reviewed and approved offline

Issue #257, opened 2019-05-18
Add subjectsInfo cookie


georgiashay:
Add a cookie with subjectsInfo that can be used before it loads from the API

georgiashay:
Update: the object is way too big to be stored in a cookie

npfoss:
What about localstorage? I've heard of people dumping Vuex state there and then reloading it when the user comes back. There might even be a built in way to do that?

miriam-rittenberg:
I think it's too large for localStorage as well

npfoss:
[discussed at courseroad meeting]

resolution, for the record:
store subjectInfo with empty course descriptions (since those are big and not immediately essential) and load the full data in the background

PR #327, opened 2019-10-22
Local storage 2

miriam-rittenberg:
This should store subject_id, title, offered_spring, offered_fall, and offered_iap in local storage, so that we can load part of subjectsInfo before we get it from FireRoad. I didn't notice a speed increase when testing, but hopefully this will help if FireRoad is taking a while to load.


georgiashay:
I may be misunderstanding the original issue, but I thought we were going to put everything in the Vuex store into localStorage (minus some things from subjectsInfo that were too big).  Was that part of this issue?

miriam-rittenberg:
I wasn't sure if I should put the whole store in, because issue #257 just mentioned subjectsInfo. I can add the rest of the state though.

georgiashay:
I think it would be useful to have the entire state loaded from storage if possible.  @npfoss what do you think?

miriam-rittenberg:
Should I update local storage whenever anything changes, or just once, say when we leave MainPage?

georgiashay:
I think MainPage has an on window unloaded function?  This should run before MainPage is left and should be sufficient to do just when the user leaves MainPage.

miriam-rittenberg:
I don't see an on window unloaded function. Possibly I should just add a set local storage part to the setters

miriam-rittenberg:
The whole state is now being stored in local storage just before MainPage is unloaded, and loads it into the state during mounted.

npfoss:
I'm getting this error on page load, and selected subjects is empty:
image
(This is in Firefox)

99% sure this is because I have stuff stored in localStorage already on localhost from another app, so it should be a straightforward fix, I'm just not sure how to properly compartmentalize where different things are stored. The solution might also be "no one else is messing with localStorage on courseroad.mit.edu so just clear your site data when developing"

src/stores/courseData.js
```diff
@@ -287,28 +295,45 @@ const store = new Vuex.Store({
...
-  totalObject.in_class_hours /= matchingClasses.length;
-  totalObject.out_of_class_hours /= matchingClasses.length;
-  return totalObject;
+    );
+    const totalObject = matchingClasses.reduce(function (accumObject, nextClass) {
+      if (store.state.fullSubjectsInfo === false) {
+        return {
+          offered_spring: accumObject.offered_spring || nextClass.offered_spring,
+          offered_IAP: accumObject.offered_IAP || nextClass.offered_IAP,
+          offered_fall: accumObject.offered_fall || nextClass.offered_fall
```
npfoss: Since `undefined || undefined` is `undefined` you may not actually have to duplicate this and have the if statement. The only way I can see this being a problem is if other code checks for `offered_summer in obj` instead of `obj.offered_summer === undefined`

npfoss:
note from the meeting: deeply consider how things could go wrong if the stored data gets out of date

also we need version numbers first

src/pages/MainPage.vue
```diff
@@ -327,11 +326,22 @@ export default {
...
+      localStorage.clear();
+      this.$cookies.set('versionNumber', this.$store.state.versionNumber);
+    }
+  },
   mounted () {
     const today = new Date();
     const month = today.getMonth();
     this.$store.commit('setCurrentSemester', (month >= 4 && month <= 10) ? 1 : 3);
-
+    if (localStorage.courseRoadStore !== undefined && this.cookiesAllowed) {
```
npfoss: why do we need to check if cookies are allowed?

src/pages/MainPage.vue
```diff
@@ -327,11 +326,22 @@ export default {
       deep: true
     }
   },
+  created () {
+    if (parseInt(this.$cookies.get('versionNumber')) !== this.$store.state.versionNumber) {
+      console.log('Warning: the version number has changed.');
+      // do whatever needs to happen when the version changed, probably including clearing local storage
+      // then update the version number cookie
+      localStorage.clear();
```
npfoss: I appreciate you doing this part on a separate branch and then adding this line here, that's exactly the best practice (IMO)

PR #329, opened 2019-10-24
Custom Colors added

eralwen:
Added custom colors info to mixin and then pass whole subject into customColors function, since every custom colors are passed by something else


src/mixins/colorMixin.js
```diff
@@ -12,7 +13,14 @@ export default {
...
   methods: {
-    courseColor: function (id) {
+    // courseColor takes in subject
+    courseColor: function (subject) {
+      // Custom course have custom_color component
+      if (subject.custom_color) {
+        return 'custom_color-' + subject.custom_color.slice(1);
+      }
+      // Otherwise it's normal class which id determines color
+      let id = subject.id || subject.subject_id;
```
npfoss: elegant.

PR #334, opened 2019-11-15
User agent bug

### LINKED ISSUE #300: User agent same for different tabs
Two different tabs saving a road, observed both with user agent '... Tab 2'.

georgiashay:
Fixes #300


georgiashay:
Fixed issue but tbh this is a bandaid, right now user agent isn't ever being set on the road, it's just being passed to FireRoad on a change without updating the road.  Here I just set it on the road if there is a conflict so it can use it.

georgiashay:
I believe this is the same bug fixed in #351.

npfoss:
You're right. Merging the new changes in now, PR approval imminent

georgiashay:
Don't think it breaks anything but pushed a fix anyway

PR #338, opened 2019-11-29
Check if logged in when loading from local storage

miriam-rittenberg:


npfoss:
Also, did we ever address the issue of localstorage having undefined roads because they haven't been loaded yet? I forget...
(e.g. you log in, load one of 3 roads, then close the page. next time those two unloaded roads could potentially cause issues?)

miriam-rittenberg:
I did not figure out the undefined roads issue

PR #342, opened 2019-12-07
Road test

miriam-rittenberg:
This probably isn't ready to merge. Contains partial tests for Road and Semester. Does not contain tests for prereqs, calculating the year from the semester (bc that is done in store), or some other things. It has some of the stuff from test-auth because I merged from that a while ago, but I deleted the files that were giving lint errors. There is one lint error caused by needing to write wrapper.vm.baseYear to get baseYear to update correctly, I don't know why this happens.


npfoss:
is this ready for review? Sorry for causing all these conflicts...

(No rush though, enjoy break!)

miriam-rittenberg:
It's not ready for review yet. I'll work on it after break, or possibly during break.

miriam-rittenberg:
I finally got these tests working.

georgiashay:
We should probably add more road tests later - probably both unit and e2e.  Looks good to merge

PR #346, opened 2019-12-13
Add CGI middleware for requests to People API

ttpcodes:
This PR adds a new `CGI` handler called `people.py` that is responsible for making requests to the `People API` on behalf of any `CourseRoad` clients. Requests can be made to this endpoint with a user's `kerb` and the API will return either an error or the year of the user zero-indexed. This resolves an issue in which the credentials for interacting with the API was stored directly in the client-side code.

This PR also modifies the logic of the `webpack-dev-server` to allow for it to run its own `CGI` handler so that this logic can be tested during development as well. For this to work properly, each person will need to obtain a copy of `credentials.ini` to set up this script (which I currently have) since these credentials will not be committed to `GitHub`.

A few changes will need to be made once this is merged into `master` to make sure this works properly, including setting up any dependencies that this `CGI` middleware requires on `Scripts`. Once this is approved by someone, I'll start testing on `dev` before deploying to production.

### DIFF
--- a/build/webpack.config.dev.js
+++ b/build/webpack.config.dev.js
@@ -1,8 +1,10 @@
 'use strict'
 const webpack = require('webpack')
 const { VueLoaderPlugin } = require('vue-loader')
+const { resolve } = require('path')
 const MiniCssExtractPlugin = require('mini-css-extract-plugin')
 const HtmlWebpackPlugin = require('html-webpack-plugin')
+const cgi = require('cgi')
 
 module.exports = (env) => {
   return {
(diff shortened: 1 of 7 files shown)


npfoss:
note: we should make sure there aren't CORS errors on either side, those are starting to show up again

ttpcodes:
The request from the `Python` script to the `CourseRoad` server doesn't have a `CORS` error. As long as `CourseRoad` is able to contact the `Python` script (considering they're on the same subdomain, I'd hope no `CORS` error appears), we should be fine.

build/webpack.config.dev.js
```diff
@@ -14,6 +16,11 @@ module.exports = (env) => {
       hot: true,
       watchOptions: {
         poll: true
+      },
+      before: function (app, server, compiler) {
+        app.get('/cgi-bin/people.py', function (req, res) {
+          cgi(resolve(__dirname, '..', 'cgi-bin', 'people.py'))(req, res)
```
npfoss: what does this do? might be worth adding a comment, because it's petty opaque (to me at least)

cgi-bin/people.py
```diff
@@ -0,0 +1,48 @@
...
+                else:
+                    year = int(year)
+                    year = year - 1
+                    output = {"year": year}
+                    print(content)
+                    print(dumps(output))
+    except Exception:
+        output = {"error": "could not read credentials"}
+        print(error + content)
+        print(dumps(output))
```
npfoss, line 48: these two lines appear in every case, I think. can't we just put them once at the end?

PR #350, opened 2019-12-13
Fix Vue production builds

ttpcodes:
This PR fixes the `Webpack` configuration so that anytime a build is produced for either the public `development` site or the `production` site, Vue's mode is properly set to `production` to hide any debugging information. This fixes an issue in which all builds would always set Vue to run in `production` mode.


georgiashay:
Seems like you got rid of NODE_ENV, is that going to have an impact on anything or is everything still the same?

ttpcodes:
So it turns out that while `NODE_ENV` is what determines whether `Vue` is in `production` or `development` mode, it is automatically set by the `--mode` parameter that we have set up for the build commands in `package.json`. Us setting it ourselves was what broke `Webpack's` handling of setting the `environment` and resulted in the build always defaulting to `development` mode. 

Do let me know if anything else looks to be broken though, and I can take a look.

PR #353, opened 2019-12-15
Define types for the props

miriam-rittenberg:


georgiashay:
Awesome!  Most lint errors are gone.  It does appear to add one though:

`warning  Prop 'conflictInfo' requires default value to be set`

I'm assuming you could do a similar thing to what you did in Auth for conflictInfo?

miriam-rittenberg:
I tried adding a default value, but it started trying to look for properties of the default, which was just an empty object I think, and giving undefined errors. There's probably a way to fix it.

georgiashay:
Took a look at it and it needed to be defaulted to undefined based on the way I set everything up it is expecting either undefined, or to have all of the correct properties.

PR #354, opened 2019-12-15
Hass Electives

georgiashay:
Adds hass elective to generic courses and filters.

Still investigating why it doesn't work in the audit, might be a fireroad problem


georgiashay:
I think the remaining issue is with fireroad-server.  I can make an issue there if I can't get my fork working to fix it myself

npfoss:
Looks like this happened before the filter refactor went through. Could you update in light of that please? Sorry I took so long to get to this.

PR #355, opened 2019-12-17
Implement history mode on vue-router

### LINKED ISSUE #312: make urls prettier
i.e. improve Vue router. very opened ended

will probably also resolve #236

ttpcodes:
This PR makes various modifications to the routing to improve how URLs are handled in `CourseRoad`:
- History mode is now used for the routing, effectively removing the need for `#` characters in the URL. Accordingly, the paths for roads are now of the format `/?road=<id>` instead of `/#/#road<id>`.
- Going backwards and forwards in the browser history both allow you to navigate between roads properly, and this logic no longer depends on `jQuery` to be operational.
- Any URLs that would return a `404` will redirect to the home page.

Accordingly, the `Webpack` configuration handles history mode properly, so that server side `404` errors will pass off handling of the error over to `CourseRoad` and its routing logic.

Once this PR is approved, I'll deploy this to `dev` and make the necessary changes to `.htaccess` on `Scripts` for this to fully work before deploying to `production`.

This PR resolves #312.

### DIFF
--- a/build/webpack.config.dev.js
+++ b/build/webpack.config.dev.js
@@ -11,6 +11,7 @@ module.exports = (env) => {
       './src/app.js'
     ],
     devServer: {
+      historyApiFallback: true,
       hot: true,
       watchOptions: {
         poll: true
--- a/build/webpack.config.dev.js
+++ b/build/webpack.config.dev.js
@@ -55,6 +55,9 @@ module.exports = (env) => {
         }
       ]
     },
+    output: {
+      publicPath: env.APP_URL.indexOf('dev') !== -1 ? '/dev/' : '/'
+    },
     plugins: [
       new webpack.HotModuleReplacementPlugin(),
       new VueLoaderPlugin(),
--- a/deploy.sh
+++ b/deploy.sh
@@ -10,12 +10,12 @@ if [ "$1" = "prod" ]; then
   echo -n "You are about to deploy to the production site, are you sure? (y/n)? "
   read answer
   if [ "$answer" != "${answer#[Yy]}" ] ;then
-    scp -r dist/* $2@athena.dialup.mit.edu:/mit/courseroad/web_scripts/courseroad/
+    scp -r deploy/production/.htaccess dist/* $2@athena.dialup.mit.edu:/mit/courseroad/web_scripts/courseroad/
   else
     echo cancelled
   fi
 elif [ "$1" = "dev" ]; then
-    scp -r dist/* $2@athena.dialup.mit.edu:/mit/courseroad/web_scripts/courseroad/dev/
+    scp -r deploy/development/.htaccess dist/* $2@athena.dialup.mit.edu:/mit/courseroad/web_scripts/courseroad/dev/
 else
   echo "Invalid build location"
 fi
--- a/src/pages/MainPage.vue
+++ b/src/pages/MainPage.vue
@@ -298,7 +298,7 @@ export default {
         this.updateFulfillment(this.$store.state.fulfillmentNeeded);
       }
       if (newRoad !== '') {
-        window.history.pushState({}, this.roads[newRoad].name, './#/#road' + newRoad);
+        this.$router.push({ path: `/road/${newRoad}` });
       }
     },
     cookiesAllowed: function (newCA) {
@@ -324,6 +324,9 @@ export default {
         }
       },
       deep: true
+    },
+    $route () {
+      this.setActiveRoad();
     }
   },
   created () {
@@ -351,10 +354,6 @@ export default {
       const scrollPosition = scrollers.scrollLeft();
       borders.css({ top: 0, left: scrollWidth - 1 + scrollPosition });
     });
-
-    $(window).on('hashchange', function () {
-      this.setActiveRoad();
-    }.bind(this));
 
     this.setActiveRoad();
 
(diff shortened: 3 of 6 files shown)


georgiashay:
This is great!  

The redirect on 404 isn't really working for me, though.  When I go to some address that doesn't exist it just says Can't GET '/whatever'.  It actually does this even for valid urls for some reason, when I do localhost:8080/about, it gives that same error even though clicking on the About page goes to localhost:8080/about just fine.

One other minor thing is that the way we were differentiating local roads from saved server roads is by putting dollar signs in the road ID.  Since the dollar signs are getting escaped by the router, this looks pretty ugly as a URL.  Perhaps we could use some other character that doesn't need to be escaped?

src/pages/MainPage.vue
```diff
@@ -426,15 +425,14 @@ export default {
...
     },
     setActiveRoad: function () {
-      const roadHash = window.location.hash;
-      if (roadHash.length && roadHash.substring(0, 7) === '#/#road') {
-        const roadRequested = roadHash.substring(7);
+      if (Object.prototype.hasOwnProperty.call(this.$route.query, 'road')) {
+        const roadRequested = this.$route.query.road;
         if (roadRequested in this.roads) {
-          this.$store.commit('setActiveRoad', roadHash.substring(7));
+          this.$store.commit('setActiveRoad', roadRequested);
```
npfoss: I'm not sure the pros and cons of different ways to check for keys in an object, but was `'road' in this.$route.query` insufficient? Or simplifying even further to just one if statement:
```js
        if (this.$route.query.road in this.roads)
```
since it would be undefined if it's not there?

npfoss:
any updates on this @ttpcodes ?

ttpcodes:
The issue with accessing a `URL` two subdirectories deep not loading `CourseRoad` is probably something we shouldn't put off. If someone refreshes during normal browsing (say, while on their `Road` at `URL: https://courseroad.mit.edu/road/123`, it'll not load `CourseRoad`, and the user will have to go back to `https://courseroad.mit.edu` for it to work.

npfoss:
Oh. Yeah that is a big problem. Let's talk about it again at the meeting today

ttpcodes:
The latest commit now sets the `publicPath` variable, which should be enough to fix this problem.

deploy.sh
```diff
@@ -11,11 +11,13 @@ if [ "$1" = "prod" ]; then
   read answer
   if [ "$answer" != "${answer#[Yy]}" ] ;then
     scp -r dist/* $2@athena.dialup.mit.edu:/mit/courseroad/web_scripts/courseroad/
+    scp -r .htaccess $2@athena.dialup.mit.edu:/mit/courseroad/web_scripts/courseroad/
```
npfoss: oof this is going to take twice as many logins now, is there a way to scp two things at once?

build/webpack.config.dev.js
```diff
@@ -54,6 +57,9 @@ module.exports = (env) => {
         }
       ]
     },
+    output: {
+      publicPath: env.APP_URL.indexOf('dev') !== -1 ? '/dev/' : '/'
```
npfoss: would `env.BASE_URL` work? (here and in the `base:` place)

src/pages/MainPage.vue
```diff
@@ -426,15 +425,14 @@ export default {
...
+      if (this.roads.hasOwnProperty(this.$route.params.road)) {
+        const roadRequested = this.$route.params.road;
         if (roadRequested in this.roads) {
-          this.$store.commit('setActiveRoad', roadHash.substring(7));
+          this.$store.commit('setActiveRoad', roadRequested);
           return true;
         }
       }
-      window.location.hash = '#/#road' + this.activeRoad;
+      this.$router.push({ path: `/road/${this.activeRoad}` });
```
npfoss: I've also seen it done like
```js
router.push({ name: 'road', params: { road: this.activeRoad } })
```
for named routes, which lets us change the path name without changing all pushes to it. Seems more extensible, but idk the arguments against it.

deploy.sh
```diff
@@ -11,11 +11,13 @@ if [ "$1" = "prod" ]; then
...
   else
     echo cancelled
   fi
 elif [ "$1" = "dev" ]; then
     scp -r dist/* $2@athena.dialup.mit.edu:/mit/courseroad/web_scripts/courseroad/dev/
+    scp -r .htaccess.dev $2@athena.dialup.mit.edu:/mit/courseroad/web_scripts/courseroad/dev/
```
npfoss: will the file `.htaccess.dev' be recognized on courseroad/dev/ ? I think it needs to be renamed?

src/pages/MainPage.vue
ttpcodes: I used the method above because the `routes` we currently have aren't actually named and we couldn't use the `named route` method. If we want, we could go back and set an explicit name for each `route` which would allow us to do this.

deploy.sh
ttpcodes: Oops, thanks for catching this.

build/webpack.config.dev.js
ttpcodes: That could also work, though we'd need to update the commands in `package.json` to set `env.BASE_URL` since those don't exist right now.

deploy.sh
ttpcodes: We can `scp` two things at once provided that they're headed for the same location. This would resolve it for the `production` case but not the `development` case though since that needs to be copied as `.htaccess` (requiring a rename and counting as a separate "destination").

build/webpack.config.dev.js
npfoss: ah. nevermind then.

deploy.sh
npfoss: (resolved by naming the dev one .htaccess and putting it in a folder)

PR #356, opened 2019-12-18
Term filters

HeadMerchant:
Adds term (Fall/Spring/IAP) filters to the class search.

Closes: #347

### DIFF
--- a/src/stores/courseData.js
+++ b/src/stores/courseData.js
@@ -267,6 +267,10 @@ const store = new Vuex.Store({
   actions: {
     async loadAllSubjects ({ commit }) {
       const response = await axios.get(process.env.FIREROAD_URL + `/courses/all?full=true`);
+
+      // condense all terms offered into one attribute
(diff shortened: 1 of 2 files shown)


georgiashay:
This should maybe be branched off of #322 since it is ready for merge and has significant changes to filters.  It would also avoid adding a field to subjectsInfo.

src/stores/courseData.js
```diff
@@ -302,10 +306,11 @@ function getMatchingAttributes (gir, hass, ci) {
       offered_summer: accumObject.offered_summer || nextClass.offered_summer,
       offered_IAP: accumObject.offered_IAP || nextClass.offered_IAP,
       offered_fall: accumObject.offered_fall || nextClass.offered_fall,
+      offered_terms: offered_terms || nextClass.offered_terms,
```
georgiashay: Not sure this is doing what you want - it looks like offered_terms is a string and || will not combine them properly

npfoss:
closing in favor of #359

PR #357, opened 2019-12-18
Add semantic semester names

### LINKED ISSUE #152: semantic semester names
"Freshman Fall (2019)" etc

(leftover from #42)

georgiashay:
Closes #152 

Not 100% sure about if this makes the names too long/if this is necessary?

### DIFF
--- a/src/components/Semester.vue
+++ b/src/components/Semester.vue
@@ -14,7 +14,7 @@
                 <span slot-scope="{ hover }" :class="hover && 'hovering'" @click="openChangeYearDialog">
                   {{ semesterName }}
                   {{ semesterType }}
-                  {{ semesterYear }}
+                  <span v-if="index>0">{{ "'" + semesterYear.toString().substring(2) }}</span>
                 </span>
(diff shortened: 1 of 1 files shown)


venkatesh-sivaraman:
FWIW we discussed this for FireRoad and decided to use “1st Year Fall”, “1st Year Spring” etc. More consistent with how admins don’t like calling first-years “freshmen” for some reason, and also more encompassing of people who take gap years and the like. We can revisit that decision if you guys feel differently

georgiashay:
That would also make them shorter & more consistent length, seems like a good idea

georgiashay:
I tried it that way and I'm not sure it looks great to have 1st/2nd/3rd/etc since we already have the year

georgiashay:
We could remove the year altogether and unify with FireRoad but I think there was a reason we have the year.  Can't remember what the reason is though

npfoss:
(put my thoughts in #152 since they were more a discussion of whether the issue is even a problem)

PR #362, opened 2019-12-19
remove v-html

npfoss:
...bringing us down to 0 errors and warnings from the linter for the first time since we added it last winter! :tada:

### DIFF
--- a/src/components/Class.vue
+++ b/src/components/Class.vue
@@ -54,6 +54,7 @@
           <h3>Warnings for {{ classInfo.id }}</h3>
         </v-card-title>
         <v-card-text>
+          <!-- eslint-disable-next-line vue/no-v-html -->
           <p v-for="warning in warnings" :key="warning" v-html="warning" />
           <v-switch
             v-model="shouldOverrideWarnings"
--- a/src/components/Semester.vue
+++ b/src/components/Semester.vue
@@ -174,6 +174,9 @@ export default {
         if (subj !== undefined) {
           const semType = (this.index - 1) % 3;
 
+          // WARNING: be careful with injecting info from the subject like this
+          //  -- if we ever take user input, it could lead to XSS attacks from custom classes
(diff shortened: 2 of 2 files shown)


georgiashay:
It was originally in a v-html to prevent this:

image

Is there any way to preserve the <b> tags without using v-html?  Parse it beforehand or something?

ttpcodes:
Are all of the warnings in the form of a bolded warning and a slightly more detailed sentence following it like this? If this is the case, we'd likely want to have our warning item consist both of a title string and a details string and pass them into separate <b> and <p> tags to display it.

georgiashay:
I checked the FireRoad code and it appears that this is this case

npfoss:
Oh oops I'm just dumb sorry. Was in a hurry and thought this was the class info card one.

This isn't actually a potential vulnerability since this text only comes from FireRoad, which is definitely trusted. So I'll just have the linter ignore it instead, that's much more convenient for flexibility of displaying warnings.

georgiashay:
Wait actually we don't even get these from FireRoad

georgiashay:
It would be pretty easy to get rid of the v-html since we are apparently making them ourselves in the warnings function of Semester.vue

npfoss:
Yeah but like what if we change our minds about how we want warnings to look? With `v-html` the code would only have to change in one place. It's a really convenient way to make the child component agnostic to the format of the warnings coming from above, and doesn't actually pose a security threat, which is all the warning is upset about

georgiashay:
We are injecting some text into the warnings, from the subject ID and last offered semester.  Not sure if this could cause xss problems for custom classes

georgiashay:
Oh actually we are only doing that for subjects that exist in Fireroad's catalog, so it should be fine

npfoss:
Good catch, and good catch. I'll add a warning to our future selves

georgiashay:
Looks good to merge

PR #375, opened 2020-02-08
Enter key submit

### LINKED ISSUE #202: Audit cases where enter key should "submit form"

georgiashay:
Closes #202 

- enter on class search opens class info card with first result
- enter in major v-autocomplete already opens result once you reach the first space in result
- enter in renaming a road submits dialog
- enter in manual progress number input submits manual progress
- enter in import roads name field imports the road

### DIFF
--- a/src/components/ImportExport.vue
+++ b/src/components/ImportExport.vue
@@ -56,7 +56,7 @@
             <v-card color="red">
               <v-card-text>
                 <b>Invalid input!</b>
-                Make sure you have given this road a name, and uploaded/pasted a valid '.road' file.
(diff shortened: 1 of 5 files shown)


src/components/ImportExport.vue
```diff
@@ -156,7 +157,10 @@ export default {
     importRoad: function (event) {
       let fail = false;
       // check for legal input
-      if (this.inputtext === '' || this.roadtitle === '') {
+      if (this.inputtext === '' ||
+          this.roadtitle === '' ||
+          this.otherRoadHasName(this.roadtitle)
+      ) {
```
npfoss: I moved your check to here so the user will get some feedback instead of it just failing. Thanks for catching that bug

PR #377, opened 2020-02-08
Add support for AFS deployment

ttpcodes:
This PR updates `deploy.sh` so that it is possible to deploy over `AFS` and avoid using `SSH`.

If the script detects `aklog`, it will assume that `AFS` is installed and authenticate with `kinit` to get a `Kerberos ticket`, authenticate to the `SIPB` cell, and use `rsync` to deploy `CourseRoad`. If not installed, this script should default back to `SSH` for deployment.

### DIFF
--- a/deploy.sh
+++ b/deploy.sh
@@ -1,3 +1,4 @@
+#!/bin/bash
 # To run this script you must have 'aklog' written in ~/.bash_environment
 # (or the environment file for whatever shell you use)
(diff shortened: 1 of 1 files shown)


npfoss:
Update from the meeting: we're starting to think maybe this is adding an unnecessary amount of added complexity for two fairly specific use cases. Discuss?

PR #378, opened 2020-02-09
Expected hours tooltip

### LINKED ISSUE #224: Handle expected hours differently for quarter classes
Right now it just adds them so if you have two quarter classes that take 10 hours each it will say 20 hours even though you will have no more than 10 hours at a time.

I think we should probably check both quarters, and take the maximum total hours for each quarter.  That way if you take 2 quarter classes in the same quarter it _will_ add them together. 
 Either way it should be documented somewhere so users know what is being calculated.

georgiashay:
Closes #224 

Shows a tooltip table that breaks down expected hours by quarter and subject.  Useful to see how hard each quarter will be, and also to see where large amounts of hours come from in your schedule.

image

### DIFF
--- a/src/components/Semester.vue
+++ b/src/components/Semester.vue
@@ -27,11 +27,13 @@
               <span v-on="on">Hours: {{ semesterInformation.totalExpectedHours.toFixed(1) }}</span>
             </template>
             <div id="tooltipTable">
-              <table v-if="semesterSubjects.length" border="1">
+              <table v-if="semesterSubjects.length">
                 <tr v-if="semesterInformation.expectedHoursQuarter1.length">
                   <th v-if="semesterInformation.anyClassInSingleQuarter" rowspan="2">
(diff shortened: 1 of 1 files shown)


georgiashay:
They have [one in Vuetify 2](https://vuetifyjs.com/en/components/simple-tables), but the [one for 1.5](https://vuetifyjs.com/en/components/data-tables) doesn't seem like it can easily support merged cells + vertical headers.  

I put it in the 1.5 data table and it looks okay-ish, but I can't make it small so I honestly like the raw html one better.  If we can figure out how to make it small it would look better though.

npfoss:
Hmm I see... Alright can you just put it back to the raw html version and I'll spend some time styling it manually? Sorry for the wild goose chase, but you're right, Vuetify is worse here

georgiashay:
Looks good to me

PR #379, opened 2020-02-09
Added link to Course 6 Student Portal

### LINKED ISSUE #251: add link to course 6 audit if course 6
It would be nice if we linked the [official course 6 audit](http://eecsappsrv.mit.edu/students/) for course 6 students

HeadMerchant:
resolves #251

Computed property `isCourse6` in Audit.vue checks if any of the user's courses are in 6.

Link is rendered below the requirements tree and only if the user is a Course 6
Screen Shot 2020-02-08 at 8 11 29 PM
Screen Shot 2020-02-08 at 8 11 41 PM

### DIFF
--- a/src/components/Audit.vue
+++ b/src/components/Audit.vue
@@ -48,10 +48,13 @@
         />
       </template>
     </v-treeview>
-
-    <a v-if="isCourse6" href="http://eecsappsrv.mit.edu/students/">
-      Course 6 Student Portal (+Audit)
-    </a>
+    
+    <p v-if="isCourse6">
+      <br/>
+      <a href="http://eecsappsrv.mit.edu/students/">
+        Course 6 Student Portal (+Audit)
+      </a>
+    </p>
 
     <v-dialog v-model="progressDialog" max-width="600">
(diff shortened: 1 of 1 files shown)


miriam-rittenberg:
Does 18C count as course 6 for the purposes of who can use the student portal? What about people who minor in course 6, or do 6-7 (who currently would get the link)?

CyanoKobalamyne:
The online checklist only works for Courses 6-1, 6-2, 6-3, 6-7, 6-14, and the MEng versions of the first four. Not sure how the minor checklist works, I don't have access to that one. 6-9, 11-6, and 18C are not administered by Course 6, so the audit/checklist doesn't work with those.

miriam-rittenberg:
Thanks for the info!

georgiashay:
Maybe the link would look better at the bottom with the disclaimer text?  Or would that be too hidden?

npfoss:
No, I think it's great where it is, I just tried the obvious ways to add padding and then gave up because there are so many open PRs :P (a good problem to have)

PR #380, opened 2020-02-09
Remove unnecessary Vue.sets

### LINKED ISSUE #265: figure out which Vue.sets to remove
Vuex definitely removes the need for some, but not all of them

georgiashay:
Keep only the Vue.sets that are needed, i.e. those which cause property addition that Vue cannot detect.

Closes #265

### DIFF
--- a/src/stores/courseData.js
+++ b/src/stores/courseData.js
@@ -53,7 +53,7 @@ const store = new Vuex.Store({
   mutations: {
     addClass (state, newClass) {
       state.roads[state.activeRoad].contents.selectedSubjects[newClass.semester].push(newClass);
-      Vue.set(state.roads[state.activeRoad], 'changed', moment().format(DATE_FORMAT));
+      state.roads[state.activeRoad].changed = moment().format(DATE_FORMAT);
     },
     addFromCard (state, classItem) {
       state.addingFromCard = true;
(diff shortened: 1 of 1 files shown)


src/stores/courseData.js
```diff
@@ -260,7 +259,7 @@ const store = new Vuex.Store({
     },
     updateProgress (state, progress) {
       Vue.set(state.roads[state.activeRoad].contents.progressOverrides, progress.listID, progress.progress);
-      Vue.set(state.roads[state.activeRoad], 'changed', moment().format(DATE_FORMAT));
+      state.roads[state.activeRoad].changed = moment().format(DATE_FORMAT);
```
npfoss: I don't think I understand when set is and isn't necessary. how are these lines really different?

georgiashay: They do the same thing.  The only place that you need a Vue.set is when you are adding a property to an object that wasn't there before.  i.e. if you are adding a new road to state.roads, you need to tell Vue about it because that key wasn't there before.  But here, the 'changed' key was already there and we are just updating it.  It's already reactive so we don't have to worry about Vue.sets

georgiashay:
I ctrl+f ed the project and only found them in two files I think

src/stores/courseData.js
npfoss: Oh, I see. Thanks

PR #381, opened 2020-02-09
Added tab color and elevation

### LINKED ISSUE #318: Cleaner tabs for the roads
The tabs interface for managing multiple roads can be cleaner. Some borders and different colors with the background would be nice.

georgiashay:
Closes #318

Color debatable


npfoss:
Hmm yeah the dividers are a huge pain... the only other idea I have is to add a ` |` to the end of the names of everything before the selected tab, and a `| ` to the beginning of everything after. Not sure if that will be enough either, though

PR #382, opened 2020-02-09
Reset fullfillmentNeeded for new Roads

### LINKED ISSUE #344: When you create a new road, it has the same progress bars as the first road until you add a class

ttpcodes:
This PR resets `fulfillmentNeeded` every time we switch to a different `Road` in `CourseRoad`. This works around the issue of new `Roads` that are switched to having their `fullfillmentNeeded` being set to none and thus not having their `Fulfillment` get reset during `updateFulfillment`.

This PR closes #344.


src/pages/MainPage.vue
```diff
@@ -295,6 +295,8 @@ export default {
           _this.$store.commit('setRetrieved', newRoad);
         });
       } else if (newRoad !== '') {
+        // Reset fulfillmentNeeded so that new Roads still call FireRoad to check fulfillment.
+        this.$store.commit('resetFulfillmentNeeded');
```
georgiashay: This is calling update fulfillment twice when you create a new road while logged in (once on creation, once on name change).  This should go specifically after a new road is added instead (in addRoad on MainPage).

PR #384, opened 2020-02-09
Convert names to lowercase on Road import

ttpcodes:
This PR converts `Road` names to lowercase when verifying that the name of a `Road` in the `ImportExport` dialog doesn't conflict.

Previously, this check was ignoring case, resulting in conflicting `Roads` being creating and causing conflicts between the two `Roads`.

### DIFF
--- a/src/components/ImportExport.vue
+++ b/src/components/ImportExport.vue
@@ -231,7 +231,7 @@ export default {
     },
     otherRoadHasName: function (roadName) {
       const otherRoadNames = Object.keys(this.roads).filter(function (road) {
-        return this.roads[road].name === roadName;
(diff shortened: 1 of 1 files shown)


ttpcodes:
I'm not sure if there are any other places where this happens. I know this check already exists when verifying the `Road` name when just creating a new `Road`. It could be worthwhile to have these share the same function for checking for `Road` name conflicts via `Mixin` or something.

You should also verify this, but in my testing, `FireRoad` seems to complain about the two `Roads`, and one of them is lost on refresh.

georgiashay:
This also needs to happen when creating a new road

ttpcodes:
This check already happens on creation of a new `Road`. We just forgot to check it under `Import`.

PR #385, opened 2020-02-09
Retrieve road before creating from duplicate

### LINKED ISSUE #331: duplicating a road which hasn't been loaded yet
If you create a new road and copy one that hasn't been loaded yet (because you haven't clicked that tab), the new road is blank. There are no errors and it still works fine after you've loaded the appropriate road, so I don't think this is a big deal. The workaround for users is easy and natural--click the original tab, confirm nonempty, try again.

georgiashay:
Fixes #331

### DIFF
--- a/src/components/Auth.vue
+++ b/src/components/Auth.vue
@@ -316,7 +316,7 @@ export default {
           }
           this.$store.commit('setActiveRoad', Object.keys(this.roads)[0]);
           // Set list of unretrieved roads to all but first road ID
-          this.$store.commit('setUnretrieved', fileKeys.slice(1));
+          this.$store.commit('setUnretrieved', fileKeys);
           if (fileKeys.length) {
             return this.retrieveRoad(fileKeys[0]);
           }
--- a/src/components/Auth.vue
+++ b/src/components/Auth.vue
@@ -252,6 +252,9 @@ export default {
           road: roadData.data.file,
           ignoreSet: true
         });
+
+        _this.$store.commit('setRetrieved', roadID);
+
         _this.gettingUserData = false;
         return roadData;
       });
--- a/src/components/RoadTabs.vue
+++ b/src/components/RoadTabs.vue
@@ -167,14 +167,14 @@ export default {
         if (this.$store.state.unretrieved.indexOf(this.duplicateRoadSource) >= 0) {
           this.$emit('retrieve', this.duplicateRoadSource);
         } else {
-          this.addDialog = false;
-          this.newRoadName = '';
           this.$emit('add-road',
             this.newRoadName,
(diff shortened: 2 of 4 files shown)


src/components/Auth.vue
```diff
@@ -252,6 +252,9 @@ export default {
           road: roadData.data.file,
           ignoreSet: true
         });
+
+        _this.$store.commit('setRetrieved', roadID);
```
npfoss: this line missing seems like it should have caused more problems...

georgiashay:
sorry, am dumb

You will probably need to clear your cookies for courseroad to test this again, the buggy road won't delete itself.

georgiashay:
Do you mean road retrieval should be in the store?  I might agree but right now all of our auth/saving/retrieving related stuff is in Auth.vue, so I'm not sure if we'd want to do some sort of refactor here?

src/components/RoadTabs.vue
```diff
@@ -149,16 +161,22 @@ export default {
...
+            this.roads[this.duplicateRoadSource].contents.coursesOfStudy.slice(0),
+            this.roads[this.duplicateRoadSource].contents.selectedSubjects.map((semester) => semester.slice(0)),
+            Object.assign({}, this.roads[this.duplicateRoadSource].contents.progressOverrides)
+          );
+          this.addDialog = false;
+          this.newRoadName = '';
+        }
       }
-      this.addDialog = false;
-      this.newRoadName = '';
```
npfoss, line 161: why duplicate this at the end of both if/else branches?

src/components/RoadTabs.vue
```diff
@@ -149,16 +161,22 @@ export default {
...
-          Object.assign({}, this.roads[this.duplicateRoadSource].contents.progressOverrides)
-        );
+        if (this.$store.state.unretrieved.indexOf(this.duplicateRoadSource) >= 0) {
+          this.$emit('retrieve', this.duplicateRoadSource);
+        } else {
+          this.$emit('add-road',
+            this.newRoadName,
+            this.roads[this.duplicateRoadSource].contents.coursesOfStudy.slice(0),
+            this.roads[this.duplicateRoadSource].contents.selectedSubjects.map((semester) => semester.slice(0)),
+            Object.assign({}, this.roads[this.duplicateRoadSource].contents.progressOverrides)
```
npfoss, line 174: seems like an exact copy of the `$store.state.unretrieved` watcher. is there a nice way to combine them?

src/components/RoadTabs.vue
georgiashay: it's only in 2/3 branches; it doesn't happen if you need to retrieve the road

npfoss: oh oops, you're totally right

PR #388, opened 2020-02-18
Handle most prereqs of the form 'one subject in' or 'two subjects in' correctly

### LINKED ISSUE #368: Make prerequisites of the form "one subject in X" work correctly.

miriam-rittenberg:
Mostly resolves #368. Some of the things that still don't work are "one subject in film / music" and "one subject in probability", because I'm not sure how to tell whether a class is on probability. The others are of the form "two subject in CMS / History" or something similar, because I think making this work would require large changes to how we handle prerequisites right now. If you have ideas for handling those or for making this code involve fewer long if statements, please add to this pull request.

### DIFF
--- a/src/mixins/reqFulfillment.js
+++ b/src/mixins/reqFulfillment.js
@@ -65,8 +65,7 @@ export default {
           const req = splitReq[i];
           let idCategory;
           let numRequired;
-          // the 'ne subject' is on purpose because some say 'One' and some say 'one'
-          if (req.indexOf('ne subject in') >= 0 || req.indexOf('wo subjects in') >= 0) {
+          if (req.toLowerCase().indexOf('one subject in') >= 0) {
             if (req.indexOf('ne subject in') >= 0) {
               numRequired = 1;
             } else {
@@ -121,7 +120,7 @@ export default {
         idCategory = /CMS/;
       } else if (category === 'Literature') {
         idCategory = /21L/;
-      } else if (category === 'film' || category === 'Film') {
+      } else if (/film/i.test(category)) {
         idCategory = /film/i;
       } else if (category === 'philosophy') {
         // below 900 is Linguistics
--- a/src/mixins/reqFulfillment.js
+++ b/src/mixins/reqFulfillment.js
@@ -51,13 +51,49 @@ export default {
       reqString = reqString.replace(/''/g, '"').replace(/,[\s]+/g, ',');
       const splitReq = reqString.split(/(,|\(|\)|\/)/);
       const _this = this;
+      let skipNumber;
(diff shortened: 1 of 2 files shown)


src/mixins/reqFulfillment.js
```diff
@@ -51,13 +51,49 @@ export default {
...
           // If the string is a required prereq, it must be manually dismissed
-          splitReq[i] = 'false';
-        } else if ('()/, '.indexOf(splitReq[i]) < 0) {
+          const req = splitReq[i];
+          let idCategory;
+          let numRequired;
+          // the 'ne subject' is on purpose because some say 'One' and some say 'one'
+          if (req.indexOf('ne subject in') >= 0 || req.indexOf('wo subjects in') >= 0) {
+            if (req.indexOf('ne subject in') >= 0) {
+              numRequired = 1;
```
georgiashay: Are there any cases where it's some other number of subjects?  There are [libraries to convert words to numbers](https://github.com/finnfiddle/words-to-numbers) if we want to make this more extendable, although it's always good to keep our node_modules down.

src/mixins/reqFulfillment.js
```diff
@@ -51,13 +51,48 @@ export default {
...
+          // Set other strings (like "permission of instructor") automatically to false
           // If required as an alternative to other prereqs, will not affect fulfillment
           // If the string is a required prereq, it must be manually dismissed
-          splitReq[i] = 'false';
-        } else if ('()/, '.indexOf(splitReq[i]) < 0) {
+          const req = splitReq[i];
+          let idCategory;
+          let numRequired;
+          if (req.toLowerCase().indexOf('one subject in') >= 0) {
+            if (req.indexOf('ne subject in') >= 0) {
```
georgiashay: It would be clearer if this one was also changed to toLowerCase

miriam-rittenberg: Good point, and also did I delete the check for two subjects in? I should put that back in.

georgiashay: Unresolved

miriam-rittenberg: Which part? I changed it to 'one subject in' last week.

georgiashay: I don't see the check for "two subjects in"

miriam-rittenberg: if (lowercaseReq.indexOf('one subject in') >= 0 || lowercaseReq.indexOf('two subjects in') >= 0) {

georgiashay: Ah, must have missed that

georgiashay:
Made sure it adds the little check marks too so we can see the new feature :)

PR #393, opened 2020-02-27
Unit test filter set

georgiashay:
Unit tests for filter set 
Also moves filter code to a utilities folder so it can be accessed by the test


src/components/Road.vue
```diff
@@ -91,6 +91,19 @@ export default {
...
+      if (this.$store.state.cookiesAllowed) {
+        this.$cookies.set('visibleList' + this.roadID, JSON.stringify(newVisibleList));
+      }
+    }
+  },
+  mounted () {
+    const visibleList = JSON.parse(this.$cookies.get('visibleList' + this.roadID));
+    if (this.$store.state.cookiesAllowed && visibleList) {
+      this.visibleList = visibleList;
+    };
```
npfoss, line 106: which PR do we need to merge first? these are a totally separate feature it seems?

src/utilities/filters.js
```diff
@@ -0,0 +1,187 @@
+class Filter {
```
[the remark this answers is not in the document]
georgiashay: Pure copy paste

src/components/Road.vue
georgiashay: Yeah, this is unrelated.  Result of a bad merge after https://github.com/sipb/courseroad2/commit/314f09f76869309d828fcaf7d4dc3a53b380f117 went in to master it seems

PR #400, opened 2020-04-04
Unit test class search

georgiashay:
- Unit test class search calculates + displays results correctly when clicking filters or a text search input is given
- Added id's to the filter buttons to make them easy to click from tests


georgiashay:
Having a catalina error :/ if anyone else wants to fix the package-lock conflict it would be much appreciated until i fix it

PR #409, opened 2020-05-05
Make override warnings visual effect immediate

### LINKED ISSUE #402: override warnings doesn't happen if you don't click 'close'
If you click on the triangle and toggle the switch but then press esc to close the dialog, your change doesn't actually take effect, however if you open the dialog again, the toggle will look like you indeed switched it. (e.g. I click override warning then esc --> triangle still there --> click triangle again --> see that toggle is already on --> ???)

yimmt:
Previously, override warning effects only take place when you click a
specific button. This allows the visual effect to happen immediately at
the toggle.

resolves #402

### DIFF
--- a/src/components/Class.vue
+++ b/src/components/Class.vue
@@ -57,13 +57,14 @@
           <!-- eslint-disable-next-line vue/no-v-html -->
           <p v-for="warning in warnings" :key="warning" v-html="warning" />
           <v-switch
+            @change="$store.commit('overrideWarnings', {override:shouldOverrideWarnings,classInfo:classInfo})"
             v-model="shouldOverrideWarnings"
             label="Override warnings"
             color="orange darken-3"
           />
         </v-card-text>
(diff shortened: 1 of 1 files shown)


georgiashay:
The code looks good to me, but I can't pull to test it out as it says "unknown repository".  Did something happen to the fork? :(.  You should be able to branch the repo directly rather than making forks, let me know if you don't have access for some reason.

yimmt:
> The code looks good to me, but I can't pull to test it out as it says "unknown repository". Did something happen to the fork? :(. You should be able to branch the repo directly rather than making forks, let me know if you don't have access for some reason.



> The code looks good to me, but I can't pull to test it out as it says "unknown repository". Did something happen to the fork? :(. You should be able to branch the repo directly rather than making forks, let me know if you don't have access for some reason.

Ah sorry. It looks like I made a fork and then deleted it because I did something on the master branch that messed it up. I can branch directly but I just decided to fork to get separate the branch on my end.

PR #410, opened 2020-05-07
Toggle Optional IAP Viewing

### LINKED ISSUE #370: Option to not see IAP

tzx:
This allows users to toggle an option to not see the IAP period on their roads

resolves #370

### DIFF
--- a/src/components/Road.vue
+++ b/src/components/Road.vue
@@ -17,7 +17,7 @@
       :is-open="visibleList[index-1]"
       :adding-from-card="addingFromCard"
       :dragging-over="dragSemesterNum===index-1"
-      :hide-iap="hideIAP"
+      :hide-iap="hideIAP === 'true'"
       @open-change-year-dialog="changeYearDialog = true"
     />
     <v-dialog v-model="changeYearDialog" max-width="600">
@@ -38,6 +38,7 @@
           <v-switch
             v-model="hideIAP"
             label="Hide IAP"
+            @change="setHideIAPOption"
           />
         </v-card-text>
         <v-card-actions>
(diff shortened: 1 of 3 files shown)


src/components/Road.vue
```diff
@@ -17,6 +17,7 @@
       :is-open="visibleList[index-1]"
       :adding-from-card="addingFromCard"
       :dragging-over="dragSemesterNum===index-1"
+      :hide-iap="hideIAP === 'true'"
```
georgiashay, line 20: It would be clearer if hideIAP were a boolean to begin with

src/components/Road.vue
```diff
@@ -25,12 +26,19 @@
...
+          <h1>Road Settings</h1>
         </v-card-title>
         <v-card-text>
           <v-select
             v-model="year"
             :items="[{value: 0,text:'First Year/Freshman'},{value: 1,text:'Sophomore'},{value:2,text:'Junior'},{value:3,text:'Senior'},{value:4,text:'Super Senior'}]"
+            label="I am a..."
+            outlined
+          />
+          <v-switch
```
georgiashay, line 38: This only seems to be hiding IAP for me after a page refresh (presumably it's setting the cookie correctly, but something else is going wrong in the meantime).  Unhiding IAP does seem to work though.  This may be because it is not a boolean?

src/components/Semester.vue
```diff
@@ -1,6 +1,7 @@
 <template>
   <!-- stolen from this example: https://vuetifyjs.com/en/components/cards#grids -->
   <v-expansion-panel-content
+    v-if="hideIap ? semesterType !== 'IAP' : true"
```
georgiashay, line 4: Nitpick, but (!hideIAP || semesterType !== 'IAP') is clearer than the ternary imo.

src/components/Road.vue
yimmt: It was initially a boolean but the problem is that we can't store booleans into cookies I think. Is there a workaround for this?

georgiashay: I would just convert the string to a boolean when you get it out of the cookie/put it into a cookie.

yimmt:
Looking back at this, I think a better solution would be to use localStorage instead of cookies

georgiashay:
That seems reasonable to me

georgiashay (changes requested):
Thanks for the update!  I'm getting it to hide/unhide IAP now, but it doesn't appear to be global until after a refresh.  I.e., it's only hiding IAP for one road, and then after a refresh it will be hidden for all the roads.  

I might suggest having hideIAP live in the store, with getters/setters that reference localStorage.  If the semesters then referenced hideIAP in the store, I think they would automatically update when it changes?

tzx:
I made it so hideIAP is now in store and get its value from localStorage so it is saved when returning. It should be able to change all roads now.

I also made it so the IAP always render, since v-model is binded to the visibleList so we won't change the elements that it originally has.

Lastly, I also change the dialog name to be more fitting since we are no longer just changing the class year.

georgiashay:
Thanks!  Merging

PR #412, opened 2020-05-15
comment out most random abbreviations for shortened titles

npfoss:


georgiashay:
I feel like a lot of these would be found in class titles and some of the abbreviations would be useful.  Thoughts?

PR #413, opened 2020-05-21
Fix UI issue with placement of percentage

### LINKED ISSUE #398: percentage on hover in wrong place for additional EECS
image

probably related to changing the icon slot thing from the drag feature

MaximoMachado:
Closes #398 

From what I saw, it seemed like removing the "&nbsp" fixed the issue without causing any unwanted side effects. It's possible I missed something though so let me know if this might cause issues I'm unaware of.


MaximoMachado:
I converted the requirement container div into a flex box which necessarily keeps everything on a single line, so the percentage doesn't appear beneath the progress bar anymore. As a bonus, it also made it incredibly easy to center the left icon on some requirements, as the way I had done it previously wasn't perfectly centered.

Let me know if you catch any issues that might have cropped up from the changes. From a quick messing around with things, it seemed like everything is working as it should and looks good.

MaximoMachado:
Maybe for another issue/pull-request, I kind of dislike that currently the audit requirements will overflow off the screen with a scrollbar added. Since it can make using the buttons on some requirements near impossible, as you have to first hover over the requirement, then scroll over.

georgiashay:
Looks good, merging! Created an issue for the scroll problem as well

PR #418, opened 2020-07-21
Filter fall classes by whether they are virtual

miriam-rittenberg:


miriam-rittenberg:
What classes are you testing with? When I test with a class offered only in the spring (7.05), it doesn't show up when I use the "not virtual" filter.

georgiashay:
1.060B shows up for me

georgiashay:
I also realized that it is possible for a class to show up under virtual and not virtual. 14.121 for example, has virtual lectures and in-person recitations, so it shows up under both.  Should virtual be restricted to virtual-only?

miriam-rittenberg:
It shows up for me too, which confuses me because the notVirtual filter should match everything that has a comma followed by not a capital V in the schedule field, and that class doesn't have a schedule field

miriam-rittenberg:
> I also realized that it is possible for a class to show up under virtual and not virtual. 14.121 for example, has virtual lectures and in-person recitations, so it shows up under both. Should virtual be restricted to virtual-only?

I'm not sure

georgiashay:
I think undefined gets cast to "undefined", but that doesn't have a comma so I'm still not sure what it's matching

miriam-rittenberg:
I think it's only historical classes that still show up, is it possible FireRoad saves the schedule from the last year or something? Or are we getting the data at https://fireroad.mit.edu/reference/catalog/courses/all?full=true without alteration?

georgiashay:
I'm pretty sure we're getting the data directly from that endpoint

georgiashay:
Oh, I see the problem - fireroad-dev still has the old schedule.  It should work fine if we push to dev and use the real endpoint

miriam-rittenberg:
Then I'm not sure what's going on, or how to make it stop happening. Is there an easy way to make a filter that's an AND of two other filters?

miriam-rittenberg:
> Oh, I see the problem - fireroad-dev still has the old schedule. It should work fine if we push to dev and use the real endpoint

Oh, makes sense, I sent my most recent message before seeing this sorry

miriam-rittenberg:
> > I also realized that it is possible for a class to show up under virtual and not virtual. 14.121 for example, has virtual lectures and in-person recitations, so it shows up under both. Should virtual be restricted to virtual-only?
> 
> I'm not sure

There's probably some way to do this with a regex but I'm not sure how, do you have ideas?

miriam-rittenberg:
Maybe ([^,]*,VIRTUAL[^,]*)+

miriam-rittenberg:
the stars got turned into italics :(

georgiashay:
I was going to suggest something similar, you may need to wrap in ^ and $ because I believe right now it searches anywhere in the string.

miriam-rittenberg:
yes, you're right

venkatesh-sivaraman:
Heyo, sorry we weren’t in touch earlier! I’ve added a `virtual_status` field to the course JSON that will be empty (no schedule available) or “Virtual”, “In-Person”, or “Virtual/In-Person” depending on the presence of schedule items with locations designated virtual. You can of course use a regex but this might make it easier and remove redundant logic :)

miriam-rittenberg:
I will use your new field, Venkat, thanks! I'll be busy for the next hour or so, so I'll change it later tonight (unless Georgia thinks we should stick with the current code)

miriam-rittenberg:
Should be good now

georgiashay:
Looks good!  I think "B" standing for both might not be totally clear, not sure what to replace it with though.  Also the question mark after virtual looks out of place imo.

miriam-rittenberg:
I agree the B isn't very clear, maybe using the whole work BOTH would be better? I think it looks better with the question mark, but I'm fine with changing it.

miriam-rittenberg:
What do you think of using BOTH?

georgiashay:
LGTM

miriam-rittenberg:
Can you merge it?

PR #422, opened 2020-11-04
Override warnings immediately

### LINKED ISSUE #402: override warnings doesn't happen if you don't click 'close'

CyanoKobalamyne:
Fixes #402.
Closes #409.


georgiashay:
Looks good!

PR #424, opened 2020-12-02
Allows filter to detect GIR classes

w-nicole:
Changed regex to detect more Subjects with Prereq for GIR classes.

### DIFF
--- a/src/components/ClassInfo.vue
+++ b/src/components/ClassInfo.vue
@@ -267,12 +267,12 @@ export default {
       var thisGIRAttr = this.currentSubject.gir_attribute;
       var IDMatcher;
 
-      if(thisGIRAttr == undefined){
+      if (thisGIRAttr === undefined) {
         IDMatcher = new RegExp('(^|[^\\da-zA-Z])' + currentID.replace(/[.*+?^${}()|[\]\\]/g, '\\$&') + '(?![\\da-zA-Z])');
-      } else { //Expression taken directly from above, but modified for GIRs
-        thisGIRAttr = "GIR:"+thisGIRAttr
-        var filteredCurrentID = currentID.replace(/[.*+?^${}()|[\]\\]/g, '\\$&')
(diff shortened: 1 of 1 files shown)


src/components/ClassInfo.vue
```diff
@@ -263,7 +263,16 @@ export default {
     subjectsWithPrereq: function () {
       const currentID = this.currentSubject.subject_id;
       const currentDept = currentID.substring(0, currentID.indexOf('.'));
-      var IDMatcher = new RegExp('(^|[^\\da-zA-Z])' + currentID.replace(/[.*+?^${}()|[\]\\]/g, '\\$&') + '(?![\\da-zA-Z])');
+
+      var thisGIRAttr = this.currentSubject.gir_attribute;
+
+      if(thisGIRAttr == undefined){
+        var IDMatcher = new RegExp('(^|[^\\da-zA-Z])' + currentID.replace(/[.*+?^${}()|[\]\\]/g, '\\$&') + '(?![\\da-zA-Z])');
```
georgiashay, line 270: The IDMatcher variable is declared separately within the if and else blocks.  While this does work fine (Javascript will "hoist" the variable so it exists outside of the if statement), it is better practice/clearer to define this variable above the if statement (just "var IDMatcher;") and then assign to it within the if statement.

georgiashay (changes requested):
Everything works great!  Just one stylistic/best practices comment but ready to merge after that is addressed!

w-nicole:
I committed a new version of the code with the style changes. It's visible in Commits. Please let me know if there's anything else I should be fix, or if I didn't push the commit correctly. Thank you for all of your help!

georgiashay:
Thanks for making the changes!  I think it would be clearer to assign thisGIRAttr on one line inside the else statement (missed this the first time).  Again a minor thing, let me know if you think it is better the original way.

georgiashay:
Pushed lint to trigger Travis (just updated our testing so it doesn't stop working after the new year).  You'll need to pull these changes before working on it (git pull).  If you've already worked on it, commit your changes and then run git pull to get my updates.

w-nicole:
I changed the variable assignment to be in one line for the else statement.

georgiashay:
Everything looks good!  Merging

PR #433, opened 2021-01-12
Fix bug where the wrong requirement is removed

### LINKED ISSUE #411: Removing a requirement from the info icon removes the last requirement not the selected one

miriam-rittenberg:
Fixes #411

### DIFF
--- a/src/components/Audit.vue
+++ b/src/components/Audit.vue
@@ -420,7 +420,7 @@ export default {
       return `--percent: ${pfulfilled}%; --bar-color: ${pcolor}; --bg-color: #fff`;
     },
     deleteReq: function (req) {
-      const reqName = req['list-id'].substring(0, req['list-id'].indexOf('.reql'));
+      const reqName = req['list-id'];
       this.$store.commit('removeReq', reqName);
     },
     submitPetition: function () {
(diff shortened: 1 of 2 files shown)


src/components/Audit.vue
```diff
@@ -420,7 +420,12 @@ export default {
       return `--percent: ${pfulfilled}%; --bar-color: ${pcolor}; --bg-color: #fff`;
     },
     deleteReq: function (req) {
-      const reqName = req['list-id'].substring(0, req['list-id'].indexOf('.reql'));
+      let reqName;
+      if (req['list-id'].includes('.reql')) {
+        reqName = req['list-id'].substring(0, req['list-id'].indexOf('.reql'));
```
georgiashay, line 425: Are there any situations where it takes this branch?  I tested a bunch and it seemed like it never had the .reql

georgiashay (commented):
Looks good to me, but we can possibly remove the first branch of the if statement unless there are situations where it is taken.

src/components/Audit.vue
miriam-rittenberg: I think you're right that it never has .reql

georgiashay:
Looks good

PR #435, opened 2021-01-13
Add local fireroad mode

georgiashay:
Add mode that is local courseroad (localhost:8080) and local fireroad (localhost:8000) for help with debugging


miriam-rittenberg:
How do I run FireRoad locally? The repo has instructions for setup but not for running it.

georgiashay:
You'll need to run python manage.py makemigrations and then python manage.py migrate.  You'll need a folder for the requirements & course catalog information.  I can send you the one I'm using (it's a few years outdated but still works).  Change CATALOG_BASE_DIR in settings.py to point to that folder.  Then run update_db.py.  You might need to set environment variable DJANGO_SETTINGS_MODULE=fireroad.settings.  After that you should be able to run python manage.py runserver to start the server locally.  

I had to add a cors.py file (I can send this as well) in the fireroad directory and add 'fireroad.cors.CorsMiddleware' to the top of MIDDLEWARE_CLASSES in order to avoid cors errors when making requests from my local courseroad instance.

I also had to pin the version of ntlk to 3.4 in setup.sh, but if you're already past the setup step and that is working then it shouldn't be a problem.

miriam-rittenberg (approved):
Looks good

PR #436, opened 2021-01-14
Avoid retrieving the first road twice

### LINKED ISSUE #403: Remove duplicate calls to retrieve first road
First road is retrieved twice - once in getUserData and once when the active road is changed and since the original API call is not done yet, it shows as unretrieved and calls the API again

miriam-rittenberg:
Fixes #403


georgiashay:
Looks good! Merging

PR #376, opened 2020-02-08
Instructor search

miriam-rittenberg:

PR #437, opened 2021-01-14
Instructor search2

### LINKED ISSUE #230: Search by professor
Add searching by professor in the class search

miriam-rittenberg:
This is the version compatible with the changes to filters.
Last time we discussed adding this a couple people said it slowed down searching a lot - I don't notice any difference, but we should maybe check this some more. See #376 .
Resolves #230


georgiashay:
I don't think I'm seeing any major slowdown

georgiashay:
Looks good! Merging

PR #440, opened 2021-01-24
Fix url redirecting to default road id when it shouldn't

### LINKED ISSUE #439: Make /road/<id> open the road with that id
If you directly type in a /road/<id> url, you should be taken directly to that road once it loads and not to the first road

MaximoMachado:
Fixes #439 

It seems like the way the code is currently working, it was trying to rely on the store being initialized during the call to mounted, but that's actually not the case, it's just an empty object during mounted. 

As such, it couldn't find the road in the url and so it set the url to whatever the active road is, which is the default road. On a second call to setting the active road, it would see that the url shows the default road and that's why this issue occurred.

My current solution feels a little hacky so if you know of a way to ensure that the store is loaded when you call mounted, that'd probably be a better way to solve this issue.

### DIFF
--- a/src/pages/MainPage.vue
+++ b/src/pages/MainPage.vue
@@ -287,7 +287,6 @@ export default {
   watch: {
     // call fireroad to check fulfillment if you change active roads or change something about a road
     activeRoad: function (newRoad) {
-      this.justLoaded = false;
       if (this.$store.state.unretrieved.indexOf(newRoad) >= 0 && !this.$refs.authcomponent.gettingUserData) {
         const _this = this;
         this.$refs.authcomponent.retrieveRoad(newRoad).then(function () {
@@ -296,9 +295,12 @@ export default {
       } else if (newRoad !== '') {
         this.updateFulfillment(this.$store.state.fulfillmentNeeded);
       }
-      if (newRoad !== '') {
+      // If just loaded, store isn't loaded yet
+      // and so we can't overwrite the router just yet
+      if (newRoad !== '' && !this.justLoaded) {
         this.$router.push({ path: `/road/${newRoad}` });
       }
+      this.justLoaded = false;
     },
     cookiesAllowed: function (newCA) {
       if (newCA) {
@@ -325,6 +327,9 @@ export default {
       deep: true
     },
     $route () {
+      this.setActiveRoad();
+    },
+    justLoaded () {
       this.setActiveRoad();
     }
   },
@@ -429,7 +434,6 @@ export default {
           return true;
         }
       }
-      this.$router.push({ path: `/road/${this.activeRoad}` });
       return false;
     },
     addRoad: function (roadName, cos = ['girs'], ss = Array.from(Array(16), () => []), overrides = {}) {
--- a/src/pages/MainPage.vue
+++ b/src/pages/MainPage.vue
@@ -431,6 +431,9 @@ export default {
       if (this.roads.hasOwnProperty(this.$route.params.road)) {
         this.$store.commit('setActiveRoad', roadRequested);
         return true;
+      } else if (!this.$cookies.isKey('accessInfo')) {
+        // If user isn't logged in, and bad road id in url, then redirect to default road
+        this.$router.push({ path: '/' });
       }
       return false;
     },
--- a/src/pages/MainPage.vue
+++ b/src/pages/MainPage.vue
@@ -427,12 +427,12 @@ export default {
       }
     },
     setActiveRoad: function () {
+      const roadRequested = this.$route.params.road;
       if (this.roads.hasOwnProperty(this.$route.params.road)) {
-        const roadRequested = this.$route.params.road;
-        if (roadRequested in this.roads) {
-          this.$store.commit('setActiveRoad', roadRequested);
-          return true;
-        }
+        this.$store.commit('setActiveRoad', roadRequested);
+        return true;
+      } else if (!this.justLoaded && this.$store.state.loggedIn !== undefined && !this.$store.state.loggedIn) {
+        this.$router.push({ path: '/' });
       }
       return false;
     },
--- a/src/pages/MainPage.vue
+++ b/src/pages/MainPage.vue
@@ -431,8 +431,6 @@ export default {
       if (this.roads.hasOwnProperty(this.$route.params.road)) {
         this.$store.commit('setActiveRoad', roadRequested);
         return true;
-      } else if (!this.justLoaded && this.$store.state.loggedIn !== undefined && !this.$store.state.loggedIn) {
-        this.$router.push({ path: '/' });
       }
       return false;
     },
--- a/src/pages/MainPage.vue
+++ b/src/pages/MainPage.vue
@@ -325,12 +325,6 @@ export default {
         }
       },
       deep: true
-    },
-    $route () {
-      this.setActiveRoad();
-    },
-    justLoaded () {
-      this.setActiveRoad();
     }
   },
   created () {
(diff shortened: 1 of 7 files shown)


georgiashay (changes requested):
Thanks for the fix! 

It looks like it isn't fetching the road contents if you directly navigate to a url of a road (other than the first road).

It also doesn't reset the URL if you navigate to a road ID that doesn't exist.  

The road-fetching code is due for a refactor, so let me know if you run into any trouble implementing the changes.  Thanks!

MaximoMachado:
Thanks for catching those bugs, I'll see what I can do to fix them.

MaximoMachado:
I believe both issues you found should be fixed now. Everything loads correctly. I was wondering if it would be a good idea or not to add a test for this issue to prevent it from happening again, the logic for how to handle this correctly is sort of spread out between the MainPage and Auth code.

If I do implement some tests for it, how would I go about it? I've tried doing testing with React in the past and I had enough trouble with it that I ended up not writing any tests. Any resources you'd suggest to get comfortable with testing with Vue?

georgiashay:
Looks good! 

I think this situation would best be served by an integration test, since it involves code across multiple pages/components.  We're currently using Cypress for integration testings and you can find links to resources in the .md file in the tests/e2e directory.  The good thing is the integration testing is not really dependent on Vue, so you can basically just use css selectors to select elements and then click/drag/type etc using Cypress commands.  The links in the .md files are good resources for the types of commands you can run, and you can also look at the existing tests in the e2e folders for examples (I tried to put a decent amount of comments there).  Let me know if you get stuck anywhere!

If you'd like to learn about Vue-specific testing, there's documentation and links the the .md file under tests/unit.  Those tests deal directly with the Vue components, but like I said it's probably more reasonable to use an integration test for this case.

MaximoMachado:
How would I go about mocking logging in with the tests? Apparently the issue where the road doesn't load any of its classes when navigated to by the url only occurs if you are logged in. Locally stored roads are perfectly fine. This definitely seems like an important thing to catch in the tests though.

georgiashay:
I would take a look at tests/e2e/specs/auth.js.  You need to mock a few different things to get login to work: the fetch_token, verify, and sync routes on Fireroad; as well as setting location.href (which is how it redirects to Fireroad).  You can likely reuse the roads that those tests use as well.  

Sounds like a good idea to have a test for both logged in and not logged in, given the behavior you're seeing.

Let me know if you have any more questions!

MaximoMachado:
Sounds good, I'll look into that. What's a good way to abstract away the login logic? It sounds like it'd be incredibly useful in a test to just have a function you can call to log the user in.

georgiashay:
You can create a custom cypress command (in the commands.js file, under the support directory).  You can see the drag and drop function there as an example.  You could put all of the login logic in there and then call it in the individual tests using cy.[name of the function](args).

MaximoMachado:
Currently, I'm not exactly sure how to fix the issue of correcting the url to the default road when a unknown road id is given while not logged in. The closest I can get checks if you're not logged in and redirects the url to `/` however this breaks the fixes while logged in.

This is due to the fact that when you are automatically logged in when you come to the browser, there is a small period of time in which you are not logged in yet. This is when the check for updating the active road occurs, since it's on mounted. When that occurs, it sees that you're not logged in and will redirect you to `/` which then just shows the default road when you're logged in.

If you know of some way to uniquely identify the scenario in which you're not in the process of being logged in and are not logged in, let me know as that should fix our problem, mostly. Best case scenario I've gotten so far only redirects to `/road` instead of `/road/${roadId}`.

georgiashay:
It looks like this.loggedIn in Auth.vue gets set as soon as it recognizes there are access info cookies, and then it starts the login process.  Will checking if loggedIn is false help or is it set too late?

MaximoMachado:
Yeah, that was one of the first things I tried. Unfortunately it gets set too late. I also tried checking if the user was not logged in and the user had not just loaded, but that doesn't end up fixing anything.

georgiashay:
Maybe you could check for the existence of the accessInfo cookie? That should be there as soon as you get to the page and would indicate the user is about to be logged in.  It might be possible for the token in accessInfo to expire (I think I remember it expiring in 3 days, but there might have been changes to login that make that different), then you'd end up in a situation where you attempt to login and fail.  That probably isn't too big of a deal since it wouldn't happen often, though.

MaximoMachado:
I added a check for the accessInfo cookie and that seems to have fixed the issue. It's passing all of the tests now. Thanks for the help!

src/pages/MainPage.vue
```diff
@@ -422,14 +427,14 @@ export default {
...
-        const roadRequested = this.$route.params.road;
-        if (roadRequested in this.roads) {
-          this.$store.commit('setActiveRoad', roadRequested);
-          return true;
-        }
+        this.$store.commit('setActiveRoad', roadRequested);
+        return true;
+      } else if (!this.$cookies.isKey('accessInfo')) {
+        // If user isn't logged in, and bad road id in url, then redirect to default road
+        this.$router.push({ path: '/' });
```
georgiashay, line 436: I'm not sure if it's this line that's causing it, but there is a new change in behavior where navigating to either '/road' or '/road/[first road id]' both do the same thing.  Previously '/road' got expanded to '/road/[first road id]' so there was always a consistent URL.  Can you look into what caused this change?

georgiashay (changes requested):
Thanks so much for all of the work on these tests!  I have a few minor comments and then it should be good to merge :)

src/pages/MainPage.vue
MaximoMachado: What's a good way to know what is the first road's id so I can redirect to it? Seems like in the store, they're stored in an object not an array.

MaximoMachado:
So, I'm able to get it to redirect to `/road/$defaultroad$` however it for whatever reason breaks the first test in roads.js. I'm not sure why though, it just seems as if the test doesn't click on the specific road tab before attempting to click the edit button for the road it's trying to edit. However, if that's the issue it should've been apparent before my changes.

However, my change doesn't work if there is some other road that is the first road. If you delete `$defaultroad$` and then go to a url with a bad road id, the default road will appear back in your road tabs and you'll be shown it instead of whatever road should be the first one.

src/pages/MainPage.vue
georgiashay: I believe that Object.keys() and iteration always produce the same ordering in javascript, so you should just be able to take the first entry in Object.keys() of the roads.

georgiashay:
It seems like what is going on here is that when you create a new road, it should automatically switch to the tab for that road.  Somehow the URL changes have broken that behavior, and since it isn't on the right tab the edit button isn't visible.

MaximoMachado:
When a new road is created, how does the website navigate you to that tab? Is it through a url redirect or a state change or something else. My initial guess as to why the change to `/road/$defaultroad$` would break things, is that the road creation redirects the url and then the very first call to setting the active road can't find the new road created in the store, which redirects you back to the default road.

As a side note, is there a way to run the e2e tests with breakpoints to look at the state of the code at time?

georgiashay:
New roads are added in the addRoad function of MainPage.vue.  The tab navigation is done by setting the activeRoad to be the new road idea.  I can't imagine why the new road wouldn't already be created, since the call to setRoad that creates the road comes before the call to setActiveRoad.  

To debug the e2e tests, I would first recommend changing `it` to `it.only` on the test you want to debug so it will run only that test.  It looks like you can put cy.pause() in the test and that will act as a breakpoint?  I haven't tried it, but [here](https://docs.cypress.io/guides/guides/debugging.html#Using) is the resource describing that.

MaximoMachado:
Looks like there was some unnecessary watches that were leading to the active road updating based on the url even when it shouldn't, i.e. in the case of a new road being created.

georgiashay (approved):
Thanks for the fixes!  Looks like it's ready to merge

PR #526, opened 2024-12-31
Add "view in hydrant" button for semesters

snowmonkey18:
- see https://github.com/sipb/courseroad2/issues/523
- currently tries to view in the corresponding Hydrant semester (if that semester isn't supported by Hydrant, hopefully Hydrant can figure out something like https://github.com/sipb/hydrant/pull/112
- alternative is to always open schedules in the latest semester with the corresponding season (lmk if this seems better)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -10,11 +10,12 @@
       "license": "MIT",
       "dependencies": {
         "@fontsource/roboto": "^5.0.13",
-        "@vitejs/plugin-vue2": "^2.3.1",
+        "@vitejs/plugin-vue2": "^2.3.3",
         "axios": "^1.6.8",
         "cgi": "^0.3.1",
         "jquery": "^3.6.0",
         "moment": "^2.29.4",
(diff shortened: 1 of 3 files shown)


psvenk (commented):
Thanks for the PR! This overall looks good to me. Just one comment: what are your thoughts on removing the "view in Hydrant" button for "prior credit"? Since this doesn't correspond to any real term, it might not make sense to have the button.

(Currently, that particular button also doesn't work because the `t` query parameter is passed with the empty string rather than not passed at all, but this shouldn't be an issue following <https://github.com/sipb/hydrant/pull/112>.)

dtemkin1:
> Thanks for the PR! This overall looks good to me. Just one comment: what are your thoughts on removing the "view in Hydrant" button for "prior credit"? Since this doesn't correspond to any real term, it might not make sense to have the button.

I agree, I pushed a change to fix this

dtemkin1 (approved):
This all looks good to me! Feel free to merge :)

PR #536, opened 2026-08-14
Rebuilding CourseRoad with Vue 3, TypeScript, and Vite

majdalafrange:
Moves the planner off Vue 2 and Vuetify. The data layer is unchanged; FireRoad API calls, the .road file format, and the shape of data saved in the browser all remain byte-compatible, so existing roads will still work.

Tests move to Vitest for unit coverage and Playwright for end-to-end testing, which replaces the old Jest and Cypress.

Just as a note: for now, npm audit reports 6 advisories, all in the vite/vitest development toolchain and none of it in shipped dependencies. To clear that, it needs the vite 8 and vitest 4 majors, which I've left as a separate change later.

Run it:
npm ci
npm run dev # http://localhost:8080/
npm test
npm run test:e2e


dtemkin1:
ok i think im done being a perfectionist... I made a few visual changes but everything looks good on my end, let me know if you disagree with them @majdalafrange and we can revert :)

